PR #4, opened 2019-12-04
[WIP] chore: prometheus test harness created

alejandroEsc:
WIP

adding tests to prometheus deployment.

### DIFF
--- a/test/addons_test.go
+++ b/test/addons_test.go
@@ -50,6 +50,7 @@ var temporarilyFilteredAddons = []string{
 
 // TestAddons tests deployment of all addons in this repository
 func TestAddons(t *testing.T) {
+	t.Log("testing filtered addon deployment")
 	cluster, err := kind.NewCluster(semver.MustParse("1.15.6"))
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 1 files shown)


test/addons_test.go
```diff
@@ -65,7 +66,7 @@ func TestAddons(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	testAddons := []v1beta1.AddonInterface{}
+	testAddons := make([]v1beta1.AddonInterface,0)
```
joejulian: > idiomatic empty initialization is to use var s []string rather than s := []string{} because it actually saves a potentially redundant allocation
```suggestion
	var testAddons []v1beta1.AddonInterface
```

test/addons_test.go
```diff
@@ -90,3 +91,41 @@ func TestAddons(t *testing.T) {
...
+	if err := temp.DeployController(promCluster); err != nil {
+		t.Fatal(err)
+	}
+
+	addons, err := temp.Addons("../addons/")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	testAddons := make([]v1beta1.AddonInterface,0)
```
joejulian:
```suggestion
	var testAddons []v1beta1.AddonInterface
```

test/addons_test.go
jkoelker: TIL. I was just about to comment the opposite.

PR #9, opened 2019-12-05
[chore] Limit CI Tests to Files Changed

shaneutt:
This PR adds cleanup to make the testing groups defined in `test/groups.yaml` instead of in go code, and adds a script to isolate tests in CI only to run group tests when a relevant addon has changed.


shaneutt:
You'll see here that the only tests that ran were for the "General" group, as I only made changes to `dashboard` and that's the group dashboard is a part of.

addons/dashboard/dashboard.yaml
```diff
@@ -12,9 +12,9 @@ metadata:
     docs.kubeaddons.mesosphere.io/dashboard: "https://github.com/kubernetes/dashboard/blob/master/README.md"
     values.chart.helm.kubeaddons.mesosphere.io/dashboard: "https://raw.githubusercontent.com/helm/charts/5e7b6640dd6b566bb136fffaca1b54da392d3074/stable/kubernetes-dashboard/values.yaml"
 spec:
-  namespace: kube-system
```
shaneutt: This was changed to instigate a test of the test group containing dashboard to verify functionality.

PR #10, opened 2019-12-06
Bumped the Nvidia chart version to support GPU metrics.

Gilbert88:
This PR bumps the Nvidia chart version to 0.2.0, which contains the GPU metrics and granfana dashboard support.

It also removes the limits for GPU driver daemonset, because GPU is considered as dedicated resources and we do not want to add limited resources to those mission critical tasks like GPU driver, to avoid dependent applications being affected if the driver being OOM killed.

This PR depends on the helm chart PR https://github.com/mesosphere/charts/pull/285

Screen Shot 2019-11-26 at 3 27 32 PM
Screen Shot 2019-11-26 at 3 29 20 PM


addons/nvidia/nvidia.yaml
```diff
@@ -54,4 +51,16 @@ spec:
         initContainers:
         - name: init-wait
           image: busybox
-          command: ['sh', '-c', 'sleep 90']
+          command: ['sh', '-c', 'sleep 180']
```
joejulian: Same comment as https://github.com/mesosphere/kubeaddons-configs/pull/350#discussion_r355059770

addons/nvidia/nvidia.yaml
```diff
@@ -54,4 +51,16 @@ spec:
...
+      nvidia-dcgm-exporter:
+        enabled: true
+        namespace: kubeaddons
+        nodeSelector:
+          konvoy.mesosphere.com/gpu-provider: NVIDIA
+        initContainers:
+        - name: init-wait
+          image: busybox
+          command: ['sh', '-c', 'sleep 200']
```
joejulian: here, too, about sleep

addons/nvidia/nvidia.yaml
```diff
@@ -26,19 +26,16 @@ spec:
...
     repo: https://mesosphere.github.io/charts/staging
-    version: 0.1.2
+    version: 0.2.0
     values: |
       ---
       nvidia-driver:
         image:
           tag: "418.87.01-centos7"
         resources:
-          limits:
```
joejulian: same as it's clone PR:
@hectorj2f You've done a ton of work on resource parameters. Do you have any issue with this change?

addons/nvidia/nvidia.yaml
Gilbert88: replied on that thread:)

addons/nvidia/nvidia.yaml
Gilbert88: ditto

addons/nvidia/nvidia.yaml
Gilbert88: see the other thread

Gilbert88:
> This addon is currently not tested as part of our CI as it is [intentionally disabled for testing](https://github.com/mesosphere/kubernetes-base-addons/blob/master/test/groups.yaml#L64).
> 
> However, would this not be a great opportunity to remove it and get it tested? @Gilbert88 is it possible to test at least deploying the addon on a [kind](https://github.com/kubernetes-sigs/kind) cluster?

It will run on Soak. Do we want to set up GPU machines for CI? It will be a bit expensive

joejulian:
Right, no, he cannot test this in a kind cluster. That would require the Teamcity runners to have GPU and those are really expensive.

Gilbert88:
Just rebased to address conflicts

PR #13, opened 2019-12-06
bump tfa to 1.0.6

jr0d:


shaneutt:
I would normally want to stop and get the addon setup in CI, however the context is that the this addon's tests are going to be handled in a later scope due to some dependency issues that need to be resolved.

See https://jira.mesosphere.com/browse/DCOS-61664 as a follow up.

PR #17, opened 2019-12-09
Add CODEOWNERS with devx owning dispatch.

jbarrick-mesosphere:


CODEOWNERS
```diff
@@ -0,0 +1,7 @@
+# Reviewer and approver configuration for addons.
+# Implemented using GitHub's CODEOWNERS file.
+# All repos are to require TWO approvals before merge.
+# Reference: https://help.github.com/en/articles/about-code-owners
+# Order is important: later rules override preceding rules
+
```
shaneutt:
```suggestion

* @mesosphere/d2iq-ksphere 

```

shaneutt:
hey @jbarrick-mesosphere just wanted to let you know: I updated the default codeowners to me and joe as per a convo with sam: we should not be using some of the larger gh team groups it seems except for special cases.

PR #19, opened 2019-12-11
[tests][kommander][metallb] enable testing of kommander 

alejandroEsc:
This tests includes deployment of kommander and additional required tooling:

   - "metallb"
    - "opsportal"
    - "cert-manager"
    - "traefik"
    - "dex"
    - "konvoyconfig"
    - "reloader"
    - "kommander"

metallb has been updated to include configuration for a local docker deployment to get this setup to work. This should not be controversial as its (metallb) value for most deployments are manually configured or configured via konvoy.


addons/metallb/metallb.yaml
```diff
@@ -33,12 +34,20 @@ spec:
           - name: default
             protocol: layer2
             # configure addresses for your network
-            addresses: []
+            addresses:
+            - 172.17.1.200-172.17.1.250
```
shaneutt: Should it be hardcoded to this? it seems this is meant to particularly support our CI environment yes?

If so, I would recommend that we provide overrides for this in the tests, specifically in that environment rather than in the defaults for the addon.

shaneutt: @alejandroEsc see if you like https://github.com/mesosphere/kubernetes-base-addons/pull/19/commits/cf8c0d7735e28c15d65e2173bfe39c063a655029 or not, this would be my suggestion

test/groups.yaml
```diff
@@ -41,6 +41,24 @@ prometheus:
     - "opsportal"
     - "dispatch"
 
+# ------------------------------------------------------------------------------
+# Kommander
+#
+# All Kommander related addons should be tested as a part of this group
+# ------------------------------------------------------------------------------
+#    "kommander": {"opsportal", "cert-manager", "defaultstorageclass-protection", "localvolumeprovisioner", "traefik", "traefik-forward-auth", "dex", "kommander"},
```
GoelDeepak: the comment and the list don't match. Is this intentional?

alejandroEsc: definitely a NIT, sorry meant to remove that!

PR #20, opened 2019-12-12
Upgrade prometheus-operator

branden:
This corresponds to the kubeaddons-configs PR at https://github.com/mesosphere/kubeaddons-configs/pull/368.

## Testing

See the testing section in the description for https://github.com/mesosphere/charts/pull/309.


branden:
CI is failing, and this appears to be the error:

```
[Step 1/1] error: unable to read URL "https://raw.githubusercontent.com/mesosphere/kubernetes-base-addons/master/addons/kudo/kudo.yaml", server reported 404 Not Found, status code=404
```

I wonder if that's caused by #18, which renamed `kudo.yaml`. to `kudo-1.yaml`.

addons/prometheus/0.31.x/prometheus-1.yaml
```diff
@@ -105,11 +105,6 @@ spec:
                 targetPort: 5601
                 interval: 30s
         prometheusSpec:
-          image:
-            tag: v2.9.2
```
samvantran: This is what prevented your change in mesosphere/charts#309 from taking effect I'm guessing? 

But rather than delete entirely, should we still pin to a much more recent stable image? Or is that already covered since we're pointing to the chart `8.3.3`?

branden: Yeah, this was causing prometheus-operator to deploy an older version of Prometheus that couldn't handle some of the updates to the chart.

The chart's default images have been updated as part of the chart upgrade, so using defaults means we're getting the new versions. I decided to remove the tag override rather than update it, since AFAIK we no longer need to pin a specific version, and using the defaults should be easier to maintain.

branden:
> it looks like m/charts points to v0.34.0 of the prom-operator now. Should we add a new folder in this repo for v0.34.x/ with the updated prometheus-1.yaml?

Oh good catch. I wonder if we need to keep the addon for the older Prometheus version, since it won't end up in a GA release. I'll just rename the dir to `0.34.x` to be accurate and see if that breaks anything.

samvantran:
>  I'll just rename the dir to 0.34.x to be accurate and see if that breaks anything.

Sounds good. Hopefully no issues 🤞 

I wonder if in the future we'll want to provide multiple versions in case users want to use an older chart (for eg bc they can't upgrade to the latest one yet). [mesosphere/universe](https://github.com/mesosphere/universe/tree/version-3.x/repo/packages) already has a pretty good system for this.

alejandroEsc:
please rebase, should help with the test.

branden:
There's a remaining issue with https://github.com/mesosphere/kubeaddons-configs/pull/368, so I'll put the WIP label back on this PR until that's ironed out.

branden:
The plan is to merge https://github.com/mesosphere/kubeaddons-configs/pull/368 as it is and then follow up with another PR to fix the bug it introduces.

PR #23, opened 2019-12-15
[test][istio] added istio testing group

alejandroEsc:
Added a few more items to the general test, and added istio, had to set 

```
security:
       selfSigned: true
```

to get it to work.

### DIFF
--- a/test/addons_test.go
+++ b/test/addons_test.go
@@ -78,6 +78,13 @@ func TestKommanderGroup(t *testing.T) {
 	}
 }
 
+func TestIstioGroup(t *testing.T) {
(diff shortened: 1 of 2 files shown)


alejandroEsc:
CI says it passed, but doesnt seem like the test passed? 

https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=2493795&_focus=1205

@shaneutt can you help me check this?

alejandroEsc:
> lgtm, we'll probably want to move the overrides to a file; or possibly restructure groups.yaml to support adding groups with overrides specific to the test group.

Good idea!

PR #26, opened 2019-12-17
chore: bump kommander to 1.171.10

juliangieseke:
depends on https://github.com/mesosphere/charts/pull/316

### DIFF
--- a/addons/kommander/1.171.x/kommander-1.yaml
+++ b/addons/kommander/1.171.x/kommander-1.yaml
@@ -44,11 +44,11 @@ spec:
             kind: ClusterIssuer
         konvoy:
           allowUnofficialReleases: true
-        kubeaddonsConfigs:
+        kubeaddonsRepository:
           versionStrategy: mapped-kubernetes-version
           versionMap:
-            1.15.4: kommander-beta4
-            1.15.5: kommander-beta6
+            1.15.4: kommander-beta7
+            1.15.5: kommander-beta7
             1.15.6: kommander-beta7
             1.16.3: kommander-beta7
 
--- a/addons/kommander/1.171.x/kommander-1.yaml
+++ b/addons/kommander/1.171.x/kommander-1.yaml
@@ -1,0 +1,72 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: ClusterAddon
+metadata:
+  name: kommander
+  labels:
+    kubeaddons.mesosphere.io/name: kommander
+  annotations:
+    catalog.kubeaddons.mesosphere.io/addon-revision: "1.171.10-1"
+    appversion.kubeaddons.mesosphere.io/kommander: "1.171.10"
+    endpoint.kubeaddons.mesosphere.io/kommander: /ops/portal/kommander
+    values.chart.helm.kubeaddons.mesosphere.io/kommander: "https://raw.githubusercontent.com/mesosphere/charts/master/stable/kommander/values.yaml"
+    helmv2.kubeaddons.mesosphere.io/upgrade-strategy: '[{"upgradeFrom": "<=0.1.22", "strategy": "delete"}]'
+spec:
+  namespace: kommander
+  kubernetes:
+    minSupportedVersion: v1.15.6
+  requires:
+    - matchLabels:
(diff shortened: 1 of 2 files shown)


addons/kommander/1.171.x/kommander-1.yaml
```diff
@@ -6,11 +6,11 @@ metadata:
...
     kubeaddons.mesosphere.io/name: kommander
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "1.163.2-1"
-    appversion.kubeaddons.mesosphere.io/kommander: "1.163.2"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "1.171.1-1"
+    appversion.kubeaddons.mesosphere.io/kommander: "1.171.1"
     endpoint.kubeaddons.mesosphere.io/kommander: /ops/portal/kommander
     values.chart.helm.kubeaddons.mesosphere.io/kommander: "https://raw.githubusercontent.com/mesosphere/charts/master/stable/kommander/values.yaml"
-    helmv2.kubeaddons.mesosphere.io/upgrade-strategy: "[{\"upgradeFrom\": \"<=0.1.22\", \"strategy\": \"delete\"}]"
+    helmv2.kubeaddons.mesosphere.io/upgrade-strategy: '[{"upgradeFrom": "<=0.1.22", "strategy": "delete"}]'
```
juliangieseke: VSCode 🤷‍♂ 🤔

juliangieseke:
@makkes changed PR so that it adds a new version, thanks for brining this up! (I was already wondering if just replacing an version would be okay 🙈 )

juliangieseke:
oh. my bad @ folder name

@shaneutt another thing that popped up talking to @natmegs : the file referenced here refers to kommander 1.150.x, which is pretty old… any idea where this is actually used and how we can update it? https://github.com/mesosphere/kubernetes-base-addons/blob/master/addons/opsportal/1.0.x/opsportal-1.yaml#L13

addons/opsportal/1.0.x/opsportal-1.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
     catalog.kubeaddons.mesosphere.io/addon-revision: "1.0.0-1"
     appversion.kubeaddons.mesosphere.io/opsportal: "1.0.0"
     endpoint.kubeaddons.mesosphere.io/opsportal: /ops/portal/
-    values.chart.helm.kubeaddons.mesosphere.io/opsportal: "https://raw.githubusercontent.com/mesosphere/charts/4155f480571eaf82c64ddd63d3d334b1105d0591/stable/opsportal/values.yaml"
+    values.chart.helm.kubeaddons.mesosphere.io/opsportal: "https://raw.githubusercontent.com/mesosphere/charts/master/stable/opsportal/values.yaml"
```
juliangieseke: seems like this line was referring to a specific git hash instead of master (like kommander / other addons). I'd assume this is a bug? cc @shaneutt @alejandroEsc

shaneutt: Technically not a bug, it's a hack. This is intentionally set to a specific sha which points to the same version as the included application, such as to make sure the link is valid as per that version.

juliangieseke: ok, so what are we supposed to do? update to new sha once https://github.com/mesosphere/charts/pull/316 is merged? sorry for these questions. its so much new stuff to understand 🙈

shaneutt: No worries! yes. once the chart merges then link with the SHA that matches the chart version. :+1:

juliangieseke: ok, thanks for clarification 🙇

natmegs: Hm a little lost as to why we track the chart version in two places? The actual chart version ref and then this link as well? 🤔

juliangieseke: removed that part from the PR, I still fully understand whats happening and it seems like it worked before.

juliangieseke:
opening new one…

PR #31, opened 2019-12-19
gcp support and added gcpdisk-csi-driver, gcpdiskprovisioner

sebbrandt87:


sebbrandt87:
for activation of another provider without any other change? really? 😕 🤦‍♂

shaneutt:
@makkes @sebbrandt87 my thoughts on the matter:

Revisions are intended to keep a flat version history and ensure that historically deployed addons have to be patched by cutting a new revision of the addon.

Adding new data to `cloudProviders` (as opposed to changing or removing existing ones) will have a `nil` effect on the deployment, lifecycle, and cleanup of the addon as the `cloudProviders` field actually operates more like metadata (in that it's used prior to deployment to make decisions about deploying the addon but does not effect operations).

From a purely functional perspective it should be entirely safe to _add a new cloudProvider_ onto a previously existing addon as is done here, but I believe this highlights two issues that we should take into account:

1. We should consider removing the `CloudProviders` field from the API in future versions (and perhaps moving it to metadata
2. We should update the documentation to give more clear instructions of precisely when a new revision is needed with the `v1beta1` versions of our API, given the context herein

I've created the following JIRAs to address these issues:

https://jira.mesosphere.com/browse/DCOS-62437
https://jira.mesosphere.com/browse/DCOS-62438

And I'll follow up on improving the documentation today.

My suggestion for this PR would be to not worry about revising addons where cloudprovider support being added is the only change.

sebbrandt87:
👍 @shaneutt @joejulian 
what about the failing tests here?
is it an overall issue?

makkes:
I basically agree with @mesosphere/mesosphere-orchestration 

However: I was approaching this PR from my standpoint of federating default Addons. If the modified Addons were part of the set of default Addons those affected Addons are ought turn up on a customer's cluster running on GCP automatically. If we don't bump the revision the `AddonsDeployment` will not be modified hence no re-deployment/re-scan of default Addons will be triggered.

As what I'm talking about is future work (rather in-progress work atm) we can decide this in another PR but still we have to consider the case.

shaneutt:
I will investigate CI @sebbrandt87 to see if we can get this unblocked.

shaneutt:
I think the CI issue here is a timing issue with Kommander, where it's not finishing before the timeout. @alejandroEsc you ran into some issues with this recently too, did you not?

sebbrandt87:
I will be out for the rest of the year from today evening on.
Please keep this branch as I will release another GCP enabled konvoy build today.

shaneutt:
Hi @sebbrandt87 just checking in, anything you need in order to help get this PR into `ready` state? let us know how we can help!

sebbrandt87:
it is ready @shaneutt , never changed since 28days ago it was already.
it was all about the ci job only

shaneutt:
Let us know if you need any assistance on this one to get unblocked!

PR #32, opened 2019-12-19
[docs] Add revisions caveats

shaneutt:
Related to https://github.com/mesosphere/kubernetes-base-addons/pull/31 this expands documentation of addon revision.


README.md
```diff
@@ -22,3 +22,11 @@ You will find that any particular addon directory (e.g. `addons/prometheus`) may
 
 The **intention of revisions is to maintain a flat history of addon changes**. If you are making changes to any particular addon you should be making a revision of that addon as a copy of the original file with the changes made therein and the `addon-revision` version updated to reflect the new version appropriately.
 
+### notes & caveats
+
+In the `v1beta1` version our of API, there are some elements of our addon specifications that act as metadata and [may be removed in future](https://jira.mesosphere.com/browse/DCOS-62438). Here are some conditions under which *a new revision will not be needed* when updating an addon:
+
+1. If you're *only adding* a new `CloudProvider`
```
makkes: I'm not sure this is true in all cases. I'm thinking about the `AddonsDeployment` we're currently building. Let's say there's an Addon in revision 1 that's deployed as part of an `AddonsDeployment` that only has AWS as cloud provider set. This Addon will not be deployed onto Kommander-managed clusters running on Azure. If you then add the Azure provider to that Addon w/o increasing the revision, this Addon will not be deemed as updated and thus not be deployed to the Kommander-managed clusters running on Azure, which in my view is a bug.

shaneutt: I see your point. You're right then that we'll have to take the strategy of revising for every change for now.

shaneutt:
This is invalid, as there are dependencies on purely static revisions in some bodies of work now.

PR #36, opened 2019-12-19
[test][localvolumepath] testing for localvolumepath is now working

alejandroEsc:
This enables testing for localvolumepath addon

### DIFF
--- a/test/addons_test.go
+++ b/test/addons_test.go
@@ -26,7 +26,7 @@ import (
 
 const (
 	defaultKubernetesVersion = "1.15.6"
-	patchStorageClass = "{\"metadata\": {\"annotations\":{\"storageclass.kubernetes.io/is-default-class\":\"false\"}}}"
+	patchStorageClass = `{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}`
 )
 var addonTestingGroups = make(map[string][]string)
 
@@ -346,4 +346,3 @@ configInline:
        controlPlaneSecurityEnabled: true
 `,
 }
-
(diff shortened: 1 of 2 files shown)


test/addons_test.go
```diff
@@ -24,8 +24,10 @@ import (
 	"github.com/mesosphere/kubeaddons/pkg/test/cluster/kind"
 )
 
-const defaultKubernetesVersion = "1.15.6"
-
+const (
+	defaultKubernetesVersion = "1.15.6"
+	patchStorageClass = "{\"metadata\": {\"annotations\":{\"storageclass.kubernetes.io/is-default-class\":\"false\"}}}"
```
shaneutt: Would this be better/cleaner?
```suggestion
	patchStorageClass = `{"metadata":{"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}`
```

test/addons_test.go
```diff
@@ -107,6 +114,7 @@ func testgroup(t *testing.T, groupname string) error {
 		return err
 	}
 
+
```
shaneutt:
```suggestion
```

test/groups.yaml
```diff
@@ -77,6 +77,18 @@ istio:
...
+# Local Volume Provisioner
+#
+# All Local Volume Provisioner related addons should be tested as a part of this group
+# ------------------------------------------------------------------------------
+localvolumeprovisioner:
+    - "cert-manager"
+    - "defaultstorageclass-protection"
+    - "localvolumeprovisioner"
+
+
```
shaneutt:
```suggestion
```

test/groups.yaml
```diff
@@ -77,6 +77,18 @@ istio:
...
+
```
shaneutt:
```suggestion
```

test/addons_test.go
alejandroEsc: Thought i tried that but with json issues. Ill retry, maybe removing the first quotes will work.

PR #40, opened 2019-12-19
[chore] Add v0.7.x Kudo Rev

shaneutt:
As it turns out, there are some components throughout or ecosystem that don't work properly with the current `v0.8.x` versions of Kudo, and so I'm adding this older revision to support those components until they can be migrated to the new release.


zmalik:
let's not bump back to 0.7.x as some version might not even have KUDO driver integration
cc: @nfnt

joejulian:
The kudo team won't support anything less than 0.10 at this time. Closing this in favor of finding a way to move forward.

PR #41, opened 2019-12-20
Adds local repo support and removes cert-manager deploy

jr0d:
go.mod is referencing the HEAD of kubeaddons, so this should be updated once we have a kubeadons release tag.


test/go.mod
```diff
@@ -23,3 +23,5 @@ require (
 )
 
 replace k8s.io/client-go => k8s.io/client-go v0.0.0-20191016111102-bec269661e48
+
+replace github.com/mesosphere/kubeaddons => /Users/jared/git/kubeaddons
```
alejandroEsc: I will create a release so you wont have to do this, but yeah will need to update go.mod file.

PR #42, opened 2019-12-20
traefik: remove requirement of a storage class

hectorj2f:
We don't need storage for traefik

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,3 +64,4 @@ You will find that any particular addon directory (e.g. `addons/prometheus`) may
 
 The **intention of revisions is to maintain a flat history of addon changes**. If you are making changes to any particular addon you should be making a revision of that addon as a copy of the original file with the changes made therein and the `addon-revision` version updated to reflect the new version appropriately.
 
+Right now adding revisions is a manual process (see [DCOS-62943](https://jira.mesosphere.com/browse/DCOS-62943) related to automating this in future iterations). To make it easier for reviewers to review PRs where revisions are [manually] added make the first commit in your branch a commit to ONLY copy the previous revision to the new revision file. That way the following commits can actually include your changes and will be easier to historically follow without needing to manually diff the files.
--- a/addons/traefik/1.7.x/traefik-3.yaml
+++ b/addons/traefik/1.7.x/traefik-3.yaml
@@ -1,0 +1,91 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: ClusterAddon
+metadata:
+  name: traefik
+  labels:
+    kubeaddons.mesosphere.io/name: traefik
+    kubeaddons.mesosphere.io/provides: ingresscontroller
+    # TODO: we're temporarily supporting dependency on an existing default storage class
+    # on the cluster, this hack will trigger re-queue on Addons until one exists.
+    kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: "true"
+  annotations:
+    catalog.kubeaddons.mesosphere.io/addon-revision: "1.7.2-2"
+    appversion.kubeaddons.mesosphere.io/traefik: "1.7.2"
+    endpoint.kubeaddons.mesosphere.io/traefik: "/ops/portal/traefik"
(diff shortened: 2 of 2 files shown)


GoelDeepak:
@hectorj2f  wonder why we added that in the first place? was it just a bulk update on all addons?

hectorj2f:
@GoelDeepak honestly it seems to me it was a hack decision as you can see in the PR https://github.com/mesosphere/kubeaddons-configs/pull/207/files#diff-74a5bec821af9f2e0810c143abc9c059R11.

hectorj2f:
I have a related PR against kommander-beta8 https://github.com/mesosphere/kubernetes-base-addons/pull/43, please have a look at the PR

jaxesn:
This was Joe's comment on the original PR which added this. Is this no longer the case?

https://github.com/mesosphere/kubeaddons-configs/pull/207#discussion_r321247594

hectorj2f:
@jaxesn This is optional for traefik chart to store ACME certificates, but we don't use it.

hectorj2f:
I had a discussion with Sam I will block this PR and until a further discussion happens I will install a storage addon even if I don't need it in my case 😂

joejulian:
Since a user _could_ enable acme and require a pvc but:
- We don't have any dynamic way of requiring a storageclass only when needed
- We don't have a way to add a requirement in cluster.yaml

Eventually I'd like us to be able do the former.

hectorj2f:
In relation to this discussion, Jimmi proposed to disable traefik acme and use cert manager acme instead.

jieyu:
+1 on using cert-manager ACME so that we can potentially use another ingress in the future.

shaneutt:
@hectorj2f this has been waiting in `wip` for some time, is there anything we can do to help move this into a `ready` state? Do you need anything?

hectorj2f:
@shaneutt To move it to higher priority or if anyone can pick this task and do it. I didn't have enough time to complete the requested and agreed changes.

shaneutt:
> @shaneutt To move it to higher priority or if anyone can pick this task and do it. I didn't have enough time to complete the requested and agreed changes.

Understood, I will patch it.

shaneutt:
The testing situation doesn't have good support for what we'd want here without greatly increasing the scope of this PR, as such I've created https://jira.mesosphere.com/browse/DCOS-63330 as a follow up to expand this and related test cases.

GoelDeepak:
I have an idea of making it a little easy for the reviewers

Why don't we make two commits? 
1. Copy the old file and rename it according to the new version
2. make inline changes in the new file

That way it will be easier to see the diff and approve it

alejandroEsc:
> @jaxesn This is optional for traefik chart to store ACME certificates, but we don't use it.

could we then have a note in the chart in case we ever decide to?

jaxesn:
> I have an idea of making it a little easy for the reviewers
> 
> Why don't we make two commits?
> 
> 1. Copy the old file and rename it according to the new version
> 2. make inline changes in the new file
> 
> That way it will be easier to see the diff and approve it

I really like this idea.. I get worried with the PRs in this repo since they are full file changes, even tho maybe its only a version change.  Seems like its too easy to rubber stamp and miss something...  If this approach works, might be nice to distribute this out, or update the readme doc so others follow.

shaneutt:
@GoelDeepak & @jaxesn that sounds good to me. My recent updates reflect that, and add documentation to cover it for future contributors.

Also I just wanted to mention again that [DCOS-62943](https://jira.mesosphere.com/browse/DCOS-62943) is the follow up we can work on to get revisions into a fully automated state in the future so that developers don't even need to bother with them.

shaneutt:
> > @jaxesn This is optional for traefik chart to store ACME certificates, but we don't use it.
> 
> could we then have a note in the chart in case we ever decide to?

I would argue that this is the responsibility of the Helm chart to document so that we can make corrections to our addon if we decide to use additional features in the future, and this [is documented via the acme.persistence.enabled options in the upstream helm chart documentation](https://github.com/helm/charts/tree/master/stable/traefik).

Let me know what you think?

PR #43, opened 2019-12-20
traefik: remove requirement storage class

hectorj2f:
We don't need storage for traefik

### DIFF
--- a/addons/traefik/1.7.x/traefik-1.yaml
+++ b/addons/traefik/1.7.x/traefik-1.yaml
@@ -6,9 +6,6 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: traefik
     kubeaddons.mesosphere.io/provides: ingresscontroller
-    # TODO: we're temporarily supporting dependency on an existing default storage class
-    # on the cluster, this hack will trigger re-queue on Addons until one exists.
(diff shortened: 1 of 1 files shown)


addons/traefik/1.7.x/traefik-1.yaml
```diff
@@ -6,9 +6,6 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: traefik
     kubeaddons.mesosphere.io/provides: ingresscontroller
-    # TODO: we're temporarily supporting dependency on an existing default storage class
-    # on the cluster, this hack will trigger re-queue on Addons until one exists.
-    kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: "true"
   annotations:
     catalog.kubeaddons.mesosphere.io/addon-revision: "1.7.2-1"
```
jieyu: BTW, can you also fixing the version here. We're using traefik 1.7.12. We can do that in a separate PR.

hectorj2f:
I had a discussion with Sam I will block this PR and until a further discussion happens I will install a storage addon even if I don't need it in my case 😂

hectorj2f:
In relation to this discussion, Jimmi proposed to disable traefik acme and use cert manager acme instead.

hectorj2f:
I am closing this PR, we'll apply the changes to the new releases rather than backport it to beta-8.

PR #44, opened 2019-12-20
chore: add annotations for thanos & karma to kommander addon

natmegs:


GeorgiSTodorov:
I think we need to update the branch.

PR #47, opened 2019-12-26
fix: copy the same resource limits changed in kubeaddons-configs

hectorj2f:
we need to reuse the resource limits from https://github.com/mesosphere/kubeaddons-configs/pull/354


hectorj2f:
Before Screen Shot 2019-12-27 at 12 34 30 AM

after

PR #51, opened 2020-01-08
Add ClusterAddon for Dispatch 0.4.x

jbarrick-mesosphere:


addons/dispatch/0.4.x/dispatch-1.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - name: azure
+      enabled: false
+    - name: docker
+      enabled: false
+    - name: none
+      enabled: false
+  chartReference:
+    chart: dispatch
+    repo: https://mesosphere.github.io/dispatch
+    version: v0-4-1
```
alejandroEsc: is this version format correct?

alejandroEsc:
Please include changes for CI to test the deployment of this Item.

addons/dispatch/0.4.x/dispatch-1.yaml
jbarrick-mesosphere: For now, yes.

jbarrick-mesosphere:
> Please include changes for CI to test the deployment of this Item.

How?

jbarrick-mesosphere:
How can i troubleshoot this failure? https://teamcity.mesosphere.io/viewLog.html?buildId=2531269&buildTypeId=kubeaddons_KonvoyAddons&tab=buildLog

alejandroEsc:
> How can i troubleshoot this failure? https://teamcity.mesosphere.io/viewLog.html?buildId=2531269&buildTypeId=kubeaddons_KonvoyAddons&tab=buildLog

The test failing usually tells you that a deployment took too long to run, usually due to a failure. 

I see that dispatch is tested in the prometheus test group: https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=2531269&_focus=1300
and 
https://github.com/mesosphere/kubernetes-base-addons/blob/75d461f72ca80964a5b31e2c1cb4400e16057263/test/groups.yaml#L50

which is where this error occurred. Please test locally that the deployment does in fact succeed on a konvoy cluster.

jbarrick-mesosphere:
Do we not collect logs to make it possible to troubleshoot the test failure here?

In our tests in the Dispatch repository, we set the Minio disk to 1gb otherwise the local volume provisioner is not able to provision the PVC. Is there a way to override the disk size for this test but not for the standard deployment?

jbarrick-mesosphere:
Alright, same issue with a smaller PVC. I'm not finding instructions on how to run the test suite locally and a `go test` fails:

```
➜  test git:(dispatch-0.4.x) (⎈ jbarrick-staging-cluster:default) go test ./...
go: github.com/mesosphere/kubeaddons@v0.7.0: reading github.com/mesosphere/kubeaddons/go.mod at revision v0.7.0: unknown revision v0.7.0
➜  test git:(dispatch-0.4.x) (⎈ jbarrick-staging-cluster:default) 
```

We test the chart quite thoroughly in other environments (including kind), so I'll need to know how to either:

1. Launch the test suite locally.
2. Troubleshoot the integration test job.

alejandroEsc:
> Alright, same issue with a smaller PVC. I'm not finding instructions on how to run the test suite locally and a `go test` fails:
> 
> ```
> ➜  test git:(dispatch-0.4.x) (⎈ jbarrick-staging-cluster:default) go test ./...
> go: github.com/mesosphere/kubeaddons@v0.7.0: reading github.com/mesosphere/kubeaddons/go.mod at revision v0.7.0: unknown revision v0.7.0
> ➜  test git:(dispatch-0.4.x) (⎈ jbarrick-staging-cluster:default) 
> ```
> 
> We test the chart quite thoroughly in other environments (including kind), so I'll need to know how to either:
> 
> 1. Launch the test suite locally.
> 2. Troubleshoot the integration test job.

Please go to the test directory and run the test there:

```
cd test
go test -v  ./...
```

monitor your app when the prometheus test is run, you can get the kubeconfig file by looking at the kind cluster created here.

jbarrick-mesosphere:
That is what I showed in my last comment.

```
➜  kubernetes-base-addons git:(dispatch-0.4.x) (⎈ jbarrick-staging-cluster:default) cd test
go test -v  ./...                                                                          
go: github.com/mesosphere/kubeaddons@v0.7.0: reading github.com/mesosphere/kubeaddons/go.mod at revision v0.7.0: unknown revision v0.7.0
➜  test git:(dispatch-0.4.x) (⎈ jbarrick-staging-cluster:default) 
```

jbarrick-mesosphere:
Working with @jkoelker we figured out my GOPRIVATE setting isn't configured right. I've got it running now.

test/groups.yaml
```diff
@@ -87,6 +86,20 @@ localvolumeprovisioner:
...
 
+# ------------------------------------------------------------------------------
+# Dispatch
+#
+# All Dispatch related addons should be tested as a part of this group
+# ------------------------------------------------------------------------------
+dispatch:
+    - "cert-manager"
+    - "defaultstorageclass-protection"
+    - "localvolumeprovisioner"
```
alejandroEsc: if you are having trouble getting volumes provisioned, use the default storage path, and remove these: `"defaultstorageclass-protection" and `"localvolumeprovisioner"`

addons/dispatch/0.4.x/dispatch-1.yaml
```diff
@@ -0,0 +1,46 @@
...
+    endpoint.kubeaddons.mesosphere.io/argo-cd: /dispatch/argo-cd
+    docs.kubeaddons.mesosphere.io/argo-cd: "https://argoproj.github.io/argo-cd/user-guide/"
+    values.chart.helm.kubeaddons.mesosphere.io/dispatch: "https://raw.githubusercontent.com/mesosphere/charts/master/stable/dispatch/values.yaml"
+spec:
+  namespace: dispatch
+  kubernetes:
+    minSupportedVersion: v1.15.6
+  requires:
+    - matchLabels:
+        kubeaddons.mesosphere.io/provides: storageclass
```
alejandroEsc: @joejulian are we using this or enforcing this? I see velero doesn't but instead using a label `kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: "true"` and works as long as there is a default storage class.

alejandroEsc: I was able to confirm that with that label and removing the two volume parts dispatch was able to deploy:

```
dispatch:
    - "cert-manager"
    - "prometheus"
    - "prometheusadapter"
    - "opsportal"
    - "dispatch"
```

@joejulian @shaneutt probably should discuss this or get some advice from you, i am not exactly well versed in what approach i should suggest here.

jbarrick-mesosphere: So I added this because I ran into an issue where Dispatch deployed before the volume provisioner and the PVCs were never bound, so I'm not sure if I should remove them. I'll do a little more testing today and see if I can reproduce that issue to get more details.

jbarrick-mesosphere: If I remove the requirements and deploy in Konvoy, dispatch gets deployed before localvolumeprovisioner and I get this:

```
Name:          dispatch-kubeaddons-minio
Namespace:     dispatch
StorageClass:  
Status:        Pending
Volume:        
Labels:        app=minio
               chart=minio-2.5.14
               heritage=Tiller
               release=dispatch-kubeaddons
Annotations:   <none>
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:      
Access Modes:  
VolumeMode:    Filesystem
Mounted By:    dispatch-kubeaddons-minio-79cd5cb5f7-6tgf4
Events:
  Type    Reason         Age                  From                         Message
  ----    ------         ----                 ----                         -------
  Normal  FailedBinding  4s (x21 over 4m49s)  persistentvolume-controller  no persistent volumes available for this claim and no storage class is set
```

I'm thinking maybe my best bet is to update kubeaddons test harness to support customizing the kind cluster?

alejandroEsc: check the annotations i suggested above and retest.:

```
  annotations:
    catalog.kubeaddons.mesosphere.io/addon-revision: "0.4.1-1"
    appversion.kubeaddons.mesosphere.io/dispatch: "0.4.1"
    endpoint.kubeaddons.mesosphere.io/dispatch: /dispatch/tekton/
    docs.kubeaddons.mesosphere.io/dispatch: "https://beta-dispatch.d2iq.com/"
    appversion.kubeaddons.mesosphere.io/argo-cd: "1.2.0"
    endpoint.kubeaddons.mesosphere.io/argo-cd: /dispatch/argo-cd
    docs.kubeaddons.mesosphere.io/argo-cd: "https://argoproj.github.io/argo-cd/user-guide/"
    values.chart.helm.kubeaddons.mesosphere.io/dispatch: "https://raw.githubusercontent.com/mesosphere/charts/master/stable/dispatch/values.yaml"
    # TODO: we're temporarily supporting dependency on an existing default storage class
    # on the cluster, this hack will trigger re-queue on Addons until one exists.
    kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: "true"
```

jbarrick-mesosphere: https://github.com/kubernetes/kubernetes/issues/44370#issuecomment-343049129

> If at the time you created the PVC, you did not have a default storage class, then the admission controller will not be able to fill in the name of the default storage class into the PVC object.

So I do need the storage to exist first since I can't reliably assume a storage class.

jbarrick-mesosphere: I see @joejulian feels strongly about this :D https://github.com/kubernetes/kubernetes/issues/44370#issuecomment-528435944

alejandroEsc: > So I do need the storage to exist first since I can't reliably assume a storage class.

yep, for localvolumeprovisioner that seems to be the case. remove the requires label, and test with the annotation i mention above.

alejandroEsc: Screen Shot 2020-01-08 at 6 39 50 PM
I didnt catch the exact moment it completed, but here i show you what it looks like when it works and things are being torn down.

jbarrick-mesosphere: Right. That works in kind, but the issue I run into is on konvoy itself. It'll work here, but won't work on konvoy (at least, konvoy-in-docker).

alejandroEsc: It requires dynamic provisioning so i suspect it will work on azure and aws. We cannot yet test on a konvoy cluster per se, so we are trying to get this environment as close as possible. Your call about how to proceed.

jbarrick-mesosphere: Here's a novel solution, I guess. I can manually create the PV in the test environment:

```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: dispatch-minio-pv
spec:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 1Gi
  local:
    path: /mnt/disks/dispatch-minio-pv
  persistentVolumeReclaimPolicy: Delete
  storageClassName: localvolumeprovisioner
  volumeMode: Filesystem
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: kubernetes.io/os
          operator: In
          values:
          - linux
```

This seems like the least intrusive method that still maintains the integrity of the test.

alejandroEsc: will konvoy docker need that also to have dispatch working?

jbarrick-mesosphere: No - local volumes on konvoy-in-docker work because we attach docker volumes and mount them into /mnt/disks so that the local volume provisioner can dynamically create them.

We can do this with kubeaddons test harness, but I'll need to do some upstream work in kubeaddons to make it work (allow passing in a custom kind configuration file). I noticed we're using an older version of kubeaddons so I'm not sure how much of a pain it is to upgrade.

jbarrick-mesosphere: :boom: I added support for adding volumes to kind and everything seems to work!

joejulian: `kubeaddons.mesosphere.io/provides: storageclass` will ensure a storageclass labeled addon exists, it may not be set as default. The safer option is the  `kubeaddons.mesosphere.io/hack-requires-defaultstorageclass` _label_ as demonstrated here:
https://github.com/mesosphere/kubernetes-base-addons/blob/master/addons/traefik/1.7.x/traefik-1.yaml#L9-L11

joejulian:
note `kubeaddons.mesosphere.io/hack-requires-defaultstorageclass` is not a matchlabels require but rather a label itself. Look more closely at the reference.

jbarrick-mesosphere:
Oooh, I see now. Thanks!

test/go.mod
```diff
@@ -4,10 +4,12 @@ go 1.13
 
 require (
 	github.com/blang/semver v3.5.1+incompatible
+	github.com/docker/docker v1.4.2-0.20190916154449-92cc603036dd
 	github.com/emicklei/go-restful v2.11.1+incompatible // indirect
 	github.com/go-openapi/spec v0.19.4 // indirect
+	github.com/google/uuid v1.1.1
 	github.com/imdario/mergo v0.3.8 // indirect
-	github.com/mesosphere/kubeaddons v0.7.2
+	github.com/mesosphere/kubeaddons master
```
jbarrick-mesosphere: Can someone cut a kubeaddons release so we don't need to pin master here?

alejandroEsc: This is done, please rebase.

jbarrick-mesosphere: Thanks!

addons/dispatch/0.4.x/dispatch-1.yaml
shaneutt: This will need to be changed as it breaks things elsewhere, https://jira.mesosphere.com/browse/DCOS-62865 has been created as a follow up so that this will be fixed in CI later.

PR #54, opened 2020-01-09
prometheus: bump cpu and memory

jieyu:
Based on the observation in soak onprem (using beta8), we need to bump
the promethesus footprint. Otherwise, it will be constantly killed due
to OOM.

The increase of resources usage is likely caused by the increasing
number of default addons, resulting in more metrics being collected.

image


alejandroEsc:
We should check if we need to bumping these values changes our requirements for the number of nodes.

jieyu:
@alejandroEsc i manually verified that we don't need to increase node count for this change.

PR #62, opened 2020-01-10
chore: add 'stage' annotation to istio addon

natmegs:
Introduces a new annotation `stage.kubeaddons.mesosphere.io/<servicename>`. This PR adds the stage `Preview` to istio, but could also be used for other stages like `Beta`. This annotation value should be assigned as it is meant to be displayed.


jaxesn:
We dont have to, but if we are like this annotation approach @jbarrick-mesosphere might be something worth doing to the dispatch addons while its in beta?  We could use the word Beta instead of Preview.

PR #63, opened 2020-01-13
Add default addons.

armandgrillet:
The addons are not given yet as we don't know which ones should be installed by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ The structure of this repository follows the [Kubeaddons Catalog Documentation](
 You will find the following directories here:
 
 * `addons/` - containing the actual manifests for addon resources
+* `deployments/` - containing the default addons depending on the Kubernetes version
 * `metadata/` - containing static metadata for the addons in `addons/`
 * `test/` - containing integration tests for the addons in `addons/`
 
--- a/deployments/1.14/default-addons-deployments.yaml
+++ b/deployments/1.14/default-addons-deployments.yaml
@@ -1,0 +1,8 @@
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: AddonsDeployment
+metadata:
+  name: default-ksphere-addons-1.14.x
+  labels:
+    kubeaddons.mesosphere.io/name: default-ksphere-addons-1.14.x
+spec:
+  addons: []
--- a/deployments/1.15/default-addons-deployments.yaml
+++ b/deployments/1.15/default-addons-deployments.yaml
@@ -1,0 +1,8 @@
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: AddonsDeployment
+metadata:
+  name: default-ksphere-addons-1.15.x
+  labels:
+    kubeaddons.mesosphere.io/name: default-ksphere-addons-1.15.x
+spec:
+  addons: []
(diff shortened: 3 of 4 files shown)


deployments/1.14.x/addons.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: AddonsDeployment
+metadata:
+  name: defaultaddons-1.14.x
+  labels:
+    kubeaddons.mesosphere.io/name: defaultaddons-1.14.x
```
timaa2k: We will need at least the following here:
```
spec:
  addons:
```

deployments/1.14.x/addons.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: AddonsDeployment
+metadata:
+  name: defaultaddons-1.14.x
```
timaa2k: I would rename all of them to `default-addons-deployment`. They also would be going into the `kubeaddons` namespace.

timaa2k: The file name should probably match the name :)

armandgrillet: Would you add the Kubernetes version to the name or would you prefer all the AddonsDeployment to be called `default-addons-deployment`?

shaneutt: Would it be more appropriate to name this more precisely according to its intended application?

e.g.:

```suggestion
  name: default-ksphere-addons-1.14.x
```
Or something like that? Just wondering what you think, not a blocker.
```

File: deployments/1.14.x/default-ksphere-addons.yaml [hunk 3]
```diff
@@ -0,0 +1,8 @@
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: AddonsDeployment
+metadata:
+  name: default-ksphere-addons-1.14.x
+  labels:
+    kubeaddons.mesosphere.io/name: default-ksphere-addons-1.14.x
+spec:
+  addons:
```
makkes:
```suggestion
  addons: []
```
This has to be an empty array.

deployments/1.15.x/default-ksphere-addons.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: AddonsDeployment
+metadata:
+  name: default-ksphere-addons-1.15.x
+  labels:
+    kubeaddons.mesosphere.io/name: default-ksphere-addons-1.15.x
+spec:
+  addons:
```
makkes:
```suggestion
  addons: []
```
This has to be an empty array.

deployments/1.16.x/default-ksphere-addons.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: AddonsDeployment
+metadata:
+  name: default-ksphere-addons-1.16.x
+  labels:
+    kubeaddons.mesosphere.io/name: default-ksphere-addons-1.16.x
+spec:
+  addons:
```
makkes:
```suggestion
  addons: []
```
This has to be an empty array.

PR #65, opened 2020-01-13
Remove the KUDO addon

nfnt:
As of https://github.com/mesosphere/kubeaddons/pull/698, KUDO will be deployed as part of 'kubeaddons' instead of an addon.


shaneutt:
Please consider this blocked on the release of Konvoy 1.3 to reduce problems that can occur with dependencies with that version.

alejandroEsc:
@shaneutt is this still blocked?

shaneutt:
> @shaneutt is this still blocked?

When https://github.com/mesosphere/kubeaddons/pull/698 is corrected and merged and if it makes it into the Konvoy 1.3 release, we could consider this unblocked. Otherwise Konvoy 1.3 depends on the Kudo addon and this should hold until we've cut that release.

shaneutt:
No longer blocked, Konvoy 1.3 no longer wants to deploy this by default.

PR #69, opened 2020-01-13
[prometheus-operator] fix: upgrade path from 5.19.7 

alejandroEsc:
requires deletion of previous version first


alejandroEsc:
I haven't seen that be that case, but we are testing this upgrade path.

alejandroEsc:
> Will this delete the volume used by prometheus instance (losing metrics)?

Tested this win @branden and we saw that the PV was kept so data was not lost. Upgrade was possible and previos metrics were all still present.

PR #74, opened 2020-01-15
Kommander 1.182.14

juliangieseke:
This adds 1.182.14 version of kommander.


juliangieseke:
I missed #68 - will continue there

PR #75, opened 2020-01-15
addons/traefik: bump to 1.72.11

armandgrillet:
Also updates the kubeaddons-addon-initializer image used by the addon.


armandgrillet:
I cannot find what's wrong from the CI build logs. @shaneutt is there something I'm missing in this PR?

makkes:
You might be running into this as there's now one more revision to test so the overall time for the job exceeds 1h. As the last succeeding job took around 47min this is a possible explanation.

makkes:
fyi, I'm running the tests locally to see if they every succeed and how long they take.

PR #76, opened 2020-01-15
addons/metallb: bump metallb container images with security fixes

mhrabovcin:
JIRA(s):
- https://jira.mesosphere.com/browse/DCOS-62694
- https://jira.mesosphere.com/browse/DCOS-62733

Konvoy build:
- https://teamcity.mesosphere.io/viewLog.html?buildId=2545418&tab=queuedBuildOverviewTab

### DIFF
--- a/addons/metallb/0.8.x/metallb-1.yaml
+++ b/addons/metallb/0.8.x/metallb-1.yaml
@@ -40,10 +40,16 @@ spec:
     values: |
       ---
       controller:
+        image:
+          repository: metallb/controller
+          tag: v0.8.3
         requests:
           cpu: 100m
           memory: 100Mi
       speaker:
+        image:
+          repository: metallb/speaker
+          tag: v0.8.3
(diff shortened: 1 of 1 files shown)


shaneutt:
Just checking: are we not able to receive a chart update that would take care of this rather than tagging the components images specifically?

GoelDeepak:
> Just checking: are we not able to receive a chart update that would take care of this rather than tagging the components images specifically?

@shaneutt The way I understood is that we are using the upstream chart for metallb which is still on 0.8.1. So we have two options - 1. either to bump the new image (the way it is done in this PR) 2. or update the upstream chart which may take sometime

alejandroEsc:
@GoelDeepak  do we need this PR for GA?

GoelDeepak:
> @GoelDeepak do we need this PR for GA?

Its say security fixes so I would guess the answer would be yes

PR #79, opened 2020-01-15
dispatch: add Beta stage annotation

jieyu:
Dispatch is still in Beta.


jieyu:
I'll open another one against rc.1 branch

PR #81, opened 2020-01-16
kommander: use repository-tag as version strategy

hectorj2f:
We should rely on the kuberentes-base-addons tags to pick the tag version that matches the selected kubernetes version.

### DIFF
--- a/addons/kommander/1.163.x/kommander-1.yaml
+++ b/addons/kommander/1.163.x/kommander-1.yaml
@@ -51,12 +51,7 @@ spec:
         konvoy:
           allowUnofficialReleases: true
         kubeaddonsRepository:
-          versionStrategy: mapped-kubernetes-version
-          versionMap:
-            1.15.4: kommander-beta4
-            1.15.5: kommander-beta6
-            1.15.6: kommander-beta9
-            1.16.3: kommander-beta9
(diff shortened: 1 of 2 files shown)


addons/kommander/1.163.x/kommander-1.yaml
```diff
@@ -51,12 +51,7 @@ spec:
         konvoy:
           allowUnofficialReleases: true
         kubeaddonsRepository:
```
jimmidyson: You can safely remove this whole section actually and just stick with the chart defaults.

addons/kommander/1.173.x/kommander-1.yaml
```diff
@@ -45,11 +45,7 @@ spec:
         konvoy:
           allowUnofficialReleases: true
         kubeaddonsRepository:
```
jimmidyson: You can safely remove this whole section actually and just stick with the chart defaults.

addons/kommander/1.163.x/kommander-1.yaml
hectorj2f: Honestly I wanted to be explicit, but if that is bad, then I would remove it. I didn't find it to be a bad decision.

PR #82, opened 2020-01-16
gatekeeper: version bump

hectorj2f:
I am surprised why we didn't bump gatekeeper from 0.1.0 to 0.3.0 yet 🤔

### DIFF
--- a/addons/gatekeeper/3.0.x/gatekeeper-1.yaml
+++ b/addons/gatekeeper/3.0.x/gatekeeper-1.yaml
@@ -29,7 +29,7 @@ spec:
   chartReference:
     chart: gatekeeper
     repo: https://mesosphere.github.io/charts/staging
-    version: 0.1.0
+    version: 0.3.3
     values: |
       ---
       image:


addons/gatekeeper/3.0.x/gatekeeper-1.yaml
```diff
@@ -29,7 +29,7 @@ spec:
   chartReference:
     chart: gatekeeper
     repo: https://mesosphere.github.io/charts/staging
-    version: 0.1.0
+    version: 0.3.3
```
jimmidyson: Does this bump the version of gatekeeper? See line 11

hectorj2f: It bumps the gatekeeper chart, which has been updated but never bumped here.

dkoshkin:
Is this needed for the release tomorrow?

hectorj2f:
@dkoshkin yes, it is needed

dkoshkin:
Please open a PR against https://github.com/mesosphere/kubernetes-base-addons/tree/v1.3.0-rc.1 too.

PR #91, opened 2020-01-20
[nvidia] chore: update chart to v0.3.1

alejandroEsc:


addons/nvidia/0.3.x/nvidia-1.yaml
```diff
@@ -0,0 +1,58 @@
...
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: ClusterAddon
+metadata:
+  name: nvidia
+  namespace: kubeaddons
+  labels:
+    kubeaddons.mesosphere.io/name: nvidia
+    kubeaddons.mesosphere.io/provides: nvidia
+  annotations:
+    catalog.kubeaddons.mesosphere.io/addon-revision: "0.2.0-1"
```
shaneutt: If the appversion is 0.2.x, should we not reflect that in the directory? Currently it's under `0.3.x` which is the chart version (but not the app version I take it)?

alejandroEsc: good catch, thanks!

addons/nvidia/0.3.x/nvidia-1.yaml
```diff
@@ -0,0 +1,58 @@
...
+    version: 0.3.1
+    values: |
+      ---
+      grafana:
+        enabled: true
+      nvidia-dcgm-exporter:
+        enabled: true
+        nodeSelector:
+          konvoy.mesosphere.com/gpu-provider: NVIDIA
+      nvidia-device-plugin:
```
Gilbert88: Could you please do not remove the init containers?

See https://jira.mesosphere.com/browse/DCOS-62148 for details

PR #107, opened 2020-01-28
bump dispatch to 0.5.1

jbarrick-mesosphere:


jbarrick-mesosphere:
> We're going to be getting more strict about ensuring that new changes end up as new revisions in the coming days and weeks, just a heads up. This one can go as-is though given the context.

Good to know, thanks! I'll do it that way next time :+1:

shaneutt:
> > We're going to be getting more strict about ensuring that new changes end up as new revisions in the coming days and weeks, just a heads up. This one can go as-is though given the context.
> 
> Good to know, thanks! I'll do it that way next time +1

np, in the future we hope to automate revisions away so that the developer doesn't even have to think about them, none the less manually manage them (see [D2IQ-62943](https://jira.d2iq.com/browse/D2IQ-62943)), the goal to have contributions to addons work similarly to how Helm chart repos work.

PR #119, opened 2020-02-04
Document the release process

joejulian:
**What type of PR is this?**

Documentation

Details the release process for this repository

https://jira.d2iq.com/browse/D2IQ-63701

Please feel free to suggest changes to this process.

```release-note
Begin a regular release cadence. See RELEASE.md
```

**Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Releases are tags which are cut from `master` and other branches and are intende
 
 While you may see other tags in our [releases page](https://github.com/mesosphere/kubernetes-base-addons/releases) the only releases which are official and supported releases are those designated as non-prerelease and specifically mentioned to be official releases.
 
-**NOTE**: Do not use `master` for production, instead pick a supported release version.
+**NOTE**: Do not use `master` for production. Instead, pick a supported release version.
 
 For the release process, see the [release](RELEASE.md) document.
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,9 @@ Releases are tags which are cut from `master` and other branches and are intende
 
 While you may see other tags in our [releases page](https://github.com/mesosphere/kubernetes-base-addons/releases) the only releases which are official and supported releases are those designated as non-prerelease and specifically mentioned to be official releases.
 
-**WARNING**: Do not use `master` for production, instead pick a supported release version.
+**NOTE**: Do not use `master` for production, instead pick a supported release version.
+
+For the release process, see the [release](RELEASE.md) document.
 
 ## Creating a Release
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 This repository contains the default sets of addons needed to bootstrap [D2iQ Kubernetes](https://d2iq.com/solutions/ksphere).
 
-# Overview
+## Overview
 
 The structure of this repository follows the [Kubeaddons Catalog Documentation](https://github.com/mesosphere/kubeaddons/blob/master/tools/catalog/README.md) in reference, and uses the [Addon Revision concept covered therein](https://github.com/mesosphere/kubeaddons/blob/master/tools/catalog/README.md#special-addonrepository-options---addon-revisions).
 
@@ -13,7 +13,7 @@ You will find the following directories here:
 * `metadata/` - containing static metadata for the addons in `addons/`
 * `test/` - containing integration tests for the addons in `addons/`
 
-# Support
+## Support
 
 There is **no community support for this software at this time**, and this repository (while publicly visible) is only intended for consumption by D2iQ personnel.
 
@@ -21,7 +21,7 @@ If you are here looking for addon support as an end-user, support is given as pe
 
 Please [contact D2iQ](https://d2iq.com/contact) for questions and more information.
 
-# Releases
+## Releases
 
 Releases are tags which are cut from `master` and other branches and are intended to signify a single supported instance of `kubernetes-base-addons`.
 
@@ -31,7 +31,7 @@ While you may see other tags in our [releases page](https://github.com/mesospher
 
 For the release process, see the [release](RELEASE.md) document.
 
-## Creating a Release
+### Creating a Release
 
 Creating a releases is as simple as cutting a tag and making a [Github release](https://help.github.com/en/github/administering-a-repository/creating-releases).
 
@@ -41,6 +41,6 @@ Some supported releases are supported via the terms of support for some other KS
 
 For all other non-official releases, make sure your tag and description are distinctly different from the official release pattern, explain the purpose of your release, and mark is as a `pre-release`.
 
-# Contributing
+## Contributing
 
 See our [Contributing Documentation](CONTRIBUTING.md).
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -44,7 +44,7 @@ Each _**Thursday**_, this repository should be tagged for SOAK testing as follow
 - For the previous Kubernetes version, the last release may have been `stable-1.16-9`. The new tag for this Kubernetes version is `testing-1.16-10`.
 - The oldest supported release similarly might be `testing-1.15-27` for a prior `stable-1.15-26`.
 
-**NOTE:** If a breaking change causes a diversion from an older release of Kubernetes to a newer one, prior to tagging the older version must be branched, ie. `stable-1.16-9` would become `stable-1.16`, the changes since the `stable-1.16-9` tag would be merged into this branch, and the new _tag_ would still be `testing-1.16.10` but pointing to the last change on the `stable-1.16` branch.
+**NOTE:** If a breaking change causes a diversion from an older release of Kubernetes to a newer one, prior to tagging the older version must be branched, ie. `stable-1.16-9` would become `stable-1.16`, the changes since the `stable-1.16-9` tag would be merged into this branch, and the new _tag_ would still be `testing-1.16-10` but pointing to the last change on the `stable-1.16` branch.
 
 - This set of Addons are installed into a SOAK cluster<sup>[2](#footnote2)</sup>.
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -40,7 +40,7 @@ Each _**Thursday**_, this repository should be tagged for SOAK testing as follow
 
 - Parse the PR logs for release notes and generate and commit a Changelog.md
 - One _**tag**_ is made for the each supported version of Kubernetes with an incremented release counter
-- If the current version of Kubernetes is `1.17.2`, and the last release was `stable-1.17-2`, the new SOAK tag will be `testing-1.17-3`.
+- If the current version of Kubernetes is `1.17.2`, and the last release was `stable-1.17-5`, the new SOAK tag will be `testing-1.17-6`.
 - For the previous Kubernetes version, the last release may have been `stable-1.16-9`. The new tag for this Kubernetes version is `testing-1.16-10`.
 - The oldest supported release similarly might be `testing-1.15-27` for a prior `stable-1.15-26`.
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,61 @@
+# Release
+
+The kubernetes-base-addons repository is intended to support a minimal base set of Addons required to provide monitoring, logging, alerting, and backups for every version of Kubernetes actively supported by the upstream Kubernetes community.
+It is not intended to be tied to any specific version of Kubernetes installer (ie. Konvoy or Kommander), or Kubeaddons.
+Changes to Konvoy or Kommander to create a resource needed by an Addon should be avoided.
+
+- [Schedule](#schedule)
+- [Considerations](#considerations)
+  - [Kubernetes](#kubernetes)
+  - [Branches and Tags](#branches-and-tags)
+- [Process](#process)
+  - [Testing Release (Weekly, Thursday)](#testing-release-weekly-thursday)
+  - [Stable Release (Weekly, Wednesday)](#stable-release-weekly-wednesday)
+
+## Schedule
+
+Releases should be weekly on Wednesdays, or as needed to address CVEs.
+
+## Considerations
+
+### Kubernetes
+
+The Addons in this repo are the minimum base set of supported<sup>[1](#footnote1)</sup> Addons to be installed as a suite for any supported version of Kubernetes.
+
+Kubernetes support must be maintained for the latest general release of Kubernetes and the two prior minor releases.
+At the time of this writing, the latest Kubernetes release is 1.17.2.
+Support from this repo, at this time, should cover 1.15.x, 1.16.x, and 1.17.x.
+
+### Branches and Tags
+
+As much as possible, we will try to maintain a `master` branch that is compatible with all supported versions of Kubernetes.
+If there is a variation in the Kubernetes API which requires a _**breaking change**_ to the Addon, a _**branch**_ will be made for the prior Kubernetes versions.
+All future changes adopted into master will need to be back-ported to those branches.
+
+## Process
+
+### Testing Release (Weekly, Thursday)
+
+Each _**Thursday**_, this repository should be tagged for SOAK testing as follows:
+
+- Parse the PR logs for release notes and generate and commit a Changelog.md
+- One _**tag**_ is made for the each supported version of Kubernetes with an incremented release counter
+- If the current version of Kubernetes is `1.17.2`, and the last release was `stable-1.17-2`, the new SOAK tag will be `testing-1.17-3`.
+- For the previous Kubernetes version, the last release may have been `stable-1.16-9`. The new tag for this Kubernetes version is `testing-1.16-10`.
+- The oldest supported release similarly might be `testing-1.15-27` for a prior `stable-1.15-26`.
+
+**NOTE:** If a breaking change causes a diversion from an older release of Kubernetes to a newer one, prior to tagging the older version must be branched, ie. `stable-1.16-9` would become `stable-1.16`, the changes since the `stable-1.16-9` tag would be merged into this branch, and the new _tag_ would still be `testing-1.16.10` but pointing to the last change on the `stable-1.16` branch.
+
+- This set of Addons are installed into a SOAK cluster<sup>[2](#footnote2)</sup>.
+
+### Stable Release (Weekly, Wednesday)
+
+Each _**Wednesday**_:
+
+- As a standing agenda item in sig-ksphere-catalog, vote go/no-go on the release of the Addons that have been SOAK tested.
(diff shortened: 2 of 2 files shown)


RELEASE.md
```diff
@@ -0,0 +1,61 @@
...
+The Addons in this repo are the minimum base set of supported<sup>[1](#footnote1)</sup> Addons to be installed as a suite for any supported version of Kubernetes.
+
+Kubernetes support must be maintained for the latest general release of Kubernetes and the two prior minor releases.
+At the time of this writing, the latest Kubernetes release is 1.17.2.
+Support from this repo, at this time, should cover 1.15.x, 1.16.x, and 1.17.x.
+
+### Branches and Tags
+
+As much as possible, we will try to maintain a `master` branch that is compatible with all supported versions of Kubernetes.
+If there is a variation in the Kubernetes API which requires a _**breaking change**_ to the Addon, a _**branch**_ will be made for the prior Kubernetes versions.
```
dkoshkin: Can you clarify on what happens when some large change is introduced that's doesn't require any new Kubernetes APIs.
For example moving to Traefik 2.0.

dkoshkin: A similar scenario, when do we add new addons, what about making an addon enabled by default?

When upgrading from Konvoy 1.2 to 1.3 there were multiple new addons that were required, do we need to distinguish between new addons that are enabled by default vs new addons that are required?

dkoshkin: Are there any restrictions to what can go in build tag bump, vs a patch and minor releases?

RELEASE.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Process
+
+### Testing Release (Weekly, Thursday)
+
+Each _**Thursday**_, this repository should be tagged for SOAK testing as follows:
+
+- Parse the PR logs for release notes and generate and commit a Changelog.md
+- One _**tag**_ is made for the each supported version of Kubernetes with an incremented release counter
+- If the current version of Kubernetes is `1.17.2`, and the last release was `stable-1.17-2`, the new SOAK tag will be `testing-1.17-3`.
```
juliangieseke: I think this can be confusing, I guess`stabel-1.17-2` should not be tied to k8s `1.17.2`? maybe just change k8s version to something else in order to clarify that release counter is unrelated to k8s patch releases:
```suggestion
- If the current version of Kubernetes is `1.17.7`, and the last release was `stable-1.17-2`, the new SOAK tag will be `testing-1.17-3`.
```

RELEASE.md
```diff
@@ -0,0 +1,61 @@
...
+- For the previous Kubernetes version, the last release may have been `stable-1.16-9`. The new tag for this Kubernetes version is `testing-1.16-10`.
+- The oldest supported release similarly might be `testing-1.15-27` for a prior `stable-1.15-26`.
+
+**NOTE:** If a breaking change causes a diversion from an older release of Kubernetes to a newer one, prior to tagging the older version must be branched, ie. `stable-1.16-9` would become `stable-1.16`, the changes since the `stable-1.16-9` tag would be merged into this branch, and the new _tag_ would still be `testing-1.16.10` but pointing to the last change on the `stable-1.16` branch.
```
juliangieseke:
```suggestion
**NOTE:** If a breaking change causes a diversion from an older release of Kubernetes to a newer one, prior to tagging the older version must be branched, ie. `stable-1.16-9` would become `stable-1.16`, the changes since the `stable-1.16-9` tag would be merged into this branch, and the new _tag_ would still be `testing-1.16-10` but pointing to the last change on the `stable-1.16` branch.
```

RELEASE.md
```diff
@@ -0,0 +1,61 @@
...
+- [Considerations](#considerations)
+  - [Kubernetes](#kubernetes)
+  - [Branches and Tags](#branches-and-tags)
+- [Process](#process)
+  - [Testing Release (Weekly, Thursday)](#testing-release-weekly-thursday)
+  - [Stable Release (Weekly, Wednesday)](#stable-release-weekly-wednesday)
+
+## Schedule
+
+Releases should be weekly on Wednesdays, or as needed to address CVEs.
```
jieyu: I do think weekly is too aggressive as we need to test/soak and support all those versions. I hope all those can be automated eventually, but we're not there yet. So I suggest start slow, like bi-weekly releases or even monthly.

cc @samba @chrisgaun

RELEASE.md
jaxesn: Once this happens the first time, then that branch for the previous kubernetes version needs to be long lived and all future updates need to be "backported"?  How do we communicate that so its clear to contributors once we are in that situation?

RELEASE.md
```diff
@@ -0,0 +1,61 @@
...
+If there is a variation in the Kubernetes API which requires a _**breaking change**_ to the Addon, a _**branch**_ will be made for the prior Kubernetes versions.
+All future changes adopted into master will need to be back-ported to those branches.
+
+## Process
+
+### Testing Release (Weekly, Thursday)
+
+Each _**Thursday**_, this repository should be tagged for SOAK testing as follows:
+
+- Parse the PR logs for release notes and generate and commit a Changelog.md
```
jieyu: I do believe that we should ask the PR submitter to update a running Changelog.md, instead of doing that before the release.

RELEASE.md
jaxesn: I agree, I think weekly is probably too much. Remember konvoy "maintenance" releases (1.3.X) will happen as needed based on cops or others bug we want to ship fixes for.  1.4-beta releases will happen on a sprintly cadence.

RELEASE.md
juliangieseke: IMHO only if we find tooling that can automate it. we should try to avoid manual steps in this repo.

RELEASE.md
juliangieseke: we should keep release branches in a releasable state all the time, once this goal is accomplished, we can easily cut released whenever we need/want. IMHO this is mostly a matter of good CI on PRs. one still can (and should) do final manual tests before cuttong/publishing a release

RELEASE.md
```diff
@@ -0,0 +1,61 @@
...
+
+The Addons in this repo are the minimum base set of supported<sup>[1](#footnote1)</sup> Addons to be installed as a suite for any supported version of Kubernetes.
+
+Kubernetes support must be maintained for the latest general release of Kubernetes and the two prior minor releases.
+At the time of this writing, the latest Kubernetes release is 1.17.2.
+Support from this repo, at this time, should cover 1.15.x, 1.16.x, and 1.17.x.
+
+### Branches and Tags
+
+As much as possible, we will try to maintain a `master` branch that is compatible with all supported versions of Kubernetes.
```
makkes: Doesn't this conflict with the statement "Do not use `master` for production, instead pick a supported release version." from https://github.com/mesosphere/kubernetes-base-addons/pull/119/files#diff-04c6e90faac2675aa89e2176d2eec7d8R30?

RELEASE.md
```diff
@@ -0,0 +1,61 @@
...
+If there is a variation in the Kubernetes API which requires a _**breaking change**_ to the Addon, a _**branch**_ will be made for the prior Kubernetes versions.
+All future changes adopted into master will need to be back-ported to those branches.
```
makkes: We should probably have a section outlining how (or that) we decommission branches for versions that are EOL.

RELEASE.md
samvantran: +1 for biweekly to match the konvoy cadence

RELEASE.md
samvantran: I read this as more of a dev objective to shoot for. `master` branch will hopefully work on all k8s versions but users should still rely on the `stable-xx` tag when deploying their prod clusters.

README.md
```diff
@@ -27,7 +27,9 @@ Releases are tags which are cut from `master` and other branches and are intende
 
 While you may see other tags in our [releases page](https://github.com/mesosphere/kubernetes-base-addons/releases) the only releases which are official and supported releases are those designated as non-prerelease and specifically mentioned to be official releases.
 
-**WARNING**: Do not use `master` for production, instead pick a supported release version.
+**NOTE**: Do not use `master` for production, instead pick a supported release version.
```
samvantran:
```suggestion
**NOTE**: Do not use `master` for production, instead, pick a supported release version.
```

RELEASE.md
samvantran: > do we need to distinguish between new addons that are enabled by default vs new addons that are required?

it might make sense for us to audit all the addons and only leave the "required" ones in this repo (the ones absolutely critical for production ready kubernetes). 

Now that we can add multiple configRepos in cluster.yaml we can move out the non-essential addons to their own repos

samvantran: we'd need CI. Any new PRs created against master would also have to run tests on those long lived `stable` branches to know they work and if so either 1) auto create the backport PR 2) have a 'mergebot-like' way to create a backport PR

RELEASE.md
samvantran: I think [this line](https://github.com/mesosphere/kubernetes-base-addons/pull/119/files#diff-781a53e648f3df8d16a08ec083b04bf4R25) mentions that?

RELEASE.md
samvantran: I think @joejulian was suggesting using the same setup from PR https://github.com/mesosphere/konvoy/pull/1219 to automate all that

RELEASE.md
samvantran: +1 to julian. Since this is [written above](https://github.com/mesosphere/kubernetes-base-addons/pull/119/files#diff-781a53e648f3df8d16a08ec083b04bf4R26), I would suggest:
```suggestion
- If the current version of Kubernetes is `1.17.2`, and the last release was `stable-1.17-5`, the new SOAK tag will be `testing-1.17-6`.
```

RELEASE.md
jieyu: I don't think the release tooling solves all the problems. The people running the release process most likely does not have the context on each change, and is probably not likely to produce accurate and informative release notes as compared to the folks that write the PR.

samvantran: No no, so what PR https://github.com/mesosphere/konvoy/pull/1219 does is add tooling that enables a way to grep all release-notes from PRs. Any original author who creates a PR still has to add release-notes because, yes exactly, they are the ones with the most context on the code changes.

RELEASE.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Kubernetes
+
+The Addons in this repo are the minimum base set of supported<sup>[1](#footnote1)</sup> Addons to be installed as a suite for any supported version of Kubernetes.
+
+Kubernetes support must be maintained for the latest general release of Kubernetes and the two prior minor releases.
+At the time of this writing, the latest Kubernetes release is 1.17.2.
+Support from this repo, at this time, should cover 1.15.x, 1.16.x, and 1.17.x.
+
+### Branches and Tags
```
samvantran: Hey @joejulian what are your thoughts on having branches associated with konvoy releases? 

For example, say we have a `konvoy-1.3` branch that has all the commits up to when we GA'd 1.3. Now, for a `1.3.1` release, we can cherry-pick any commits from master that we want to include (eg security patches/bug fixes). 

This would make it easier to isolate (and track) which changes we want for a given version of Konvoy and avoid updating _all_ of the addons in the event we don't want to. Of course you can still cut tags/releases as `stable-1.15-xx` and `stable-1.16-xx` (which Konvoy would specify in its cluster.yaml configVersion) but this would simplify Konvoy releases a little more. WDYT?

README.md
joejulian:
```suggestion
**NOTE**: Do not use `master` for production. Instead, pick a supported release version.
```
As long as we're fixing grammer...

RELEASE.md
joejulian: This repo's release cadence isn't intended to match the konvoy releases. I'm ok with moving to biweekly, but I don't agree with konvoy being a reason.

The longer the release cadence, the more cognitive load is required: more changes, more discussion, more controversy, etc.

Even with the best possible suite of tests, we can’t guarantee a bug free release, but what we can do is fix those bugs quickly and get the fixes to users as soon as possible.

With a traditional slow release cycle you can still release a patch for serious issues, but less severe bugs tend to get pushed to the next release. For users experiencing those bugs, it can be very frustrating if they’re not fixed in a timely manner and can lead to an impression that the application isn’t actively maintained.

QA is another area where frequent releases actually helps. If you’re doing infrequent releases, there can be a lot of features, code and bugs added since the last release. This means that when an issue is found, it can be hard to trace its origins. On the other hand, if there have only been a relatively small amount of changes since the last release, then it makes it much easier to identify when and where a problem originated.

People also have a tendency to find bugs just before the build is due for release. With a slow release cycle this means that we generally need to delay even further until all known bugs are fixed. 

With a fast release cycle however, it’s much easier to persuade people to let the release go ahead as the bug can be fixed for the next release, which will be along soon. This allows us to use a rule of thumb is that if it’s not a regression (i.e. the bug already exists in the current release) then it shouldn’t hold up a new release. To do so would only deprive users of fixes for other issues that have been made since the last release.

RELEASE.md
joejulian: I think this creates a lot of work for developers and provides very little to no value for the customer. As long as we maintain a set of addons that successfully supports all of the kubernetes versions we support, there should be no need for a branch.

RELEASE.md
joejulian: Right, which is why I wanted to reduce that line from a scary "warning" to just a "note".

RELEASE.md
joejulian: > it might make sense for us to audit all the addons and only leave the "required" ones in this repo (the ones absolutely critical for production ready kubernetes).

I completely agree.

joejulian: For Dimitri's "similar scenario", I'll update the document.

RELEASE.md
samba: The two need not be so closely locked. We should work to maintain, and indeed enhance, the decoupling of the two. 

Konvoy needs to be able to handle upgrades of Addons without upgrading itself. Technically this is already feasible, just requires manual edits of the config file. We should simplify/automate this workflow, but that’s out of scope for this document. Main point: decoupling upgrades overall. 

Addons on a faster cadence can better facilitate discovery of issues in soak, provided soak deployment is linked to the (weekly) release process for addons. This will reduce the risk of shipping unknown bugs in Konvoy. (Naturally we need more robust testing and monitoring against soak.)

Further, as we’re leveraging these addons in other Kubernetes with Kommander, not just konvoy, the release process of Addons needs to be considered independently of Konvoy. It will effect customers on Kommander who may use mostly non-Konvoy clusters. In both Konvoy and non-Konvoy clusters, we also want to enable (optional) automatic upgrades *without upgrading konvoy*, so binding the release cadence to Konvoy is not inherently beneficial. 

Adopting a faster cycle here gets us better quality, and through automation (better than we have now) forces lower transaction cost of rapid responses, such as CVE updates.

RELEASE.md
joejulian: A bot would be best, I agree. Short of that, I don't have a good idea. The best I've got is a PR template message.

RELEASE.md
joejulian: Right, the PR is the time at which we will need to enforce that the release notes are sufficiently detailed to remove the burden from a human.

RELEASE.md
joejulian: There should (I think k8s is stable enough that we can rely on this) never be a need to tie a tag/branch to a kubernetes patch release. The api should not be changing. So stable-`major`.`minor`-`dist`. I'll add text to back this up.

RELEASE.md
GoelDeepak: We should treat `Kubernetes-base-addons` as a product by itself because (as Sam B said) it could exists independent of Konvoy. Konvoy happens to be one of its consumers. However, for release cadence, we need to also consider soak time. In the past, we settled on 2 weeks soaking to discover any bug. We either need to reduce that to 1 week or move the release cadence to 2 weeks

RELEASE.md
GoelDeepak: +1 on PR creator to also write the release notes

RELEASE.md
joejulian: Seems reasonable. The 3rd option would be to have overlapping soak clusters where testinging would be deployed every week. Two weeks later, that tag would be released while last week continues to soak.

RELEASE.md
joejulian: We can defer any additional details for now and discuss it in the sig when the problem is nearer.

RELEASE.md
samvantran: Okay got it. Thinking it over, it might be more sensible to mark the Konvoy release as 1.4.0 (or 1.4-beta) since using a newer base-addons release/tag may include new features as well as bug fixes compared to its 1.3 version. Thanks, talking this out cleared up some misunderstandings on my end.

RELEASE.md
makkes: Got it. Thanks.

juliangieseke:
I'm fine with merging this as is (as it reflects current reality), but IMHO there are a couple things that must be addressed going forward. I've pulled out a comment I was going to make on L3 into it a [gist](https://gist.github.com/juliangieseke/d8a6991e94985d7b934a1dd93d77fab9) and also posted it to slack to allow better discussion around it.

makkes:
I would like to emphasize what @joejulian said:
> The longer the release cadence, the more cognitive load is required: more changes, more discussion, more controversy, etc.
This cannot be said often enough. The more often we release, the more we will ingest the process, the less issues we will have etc.

alejandroEsc:
Please let @joejulian  merge when he is comfortable with the PR.

RELEASE.md
joejulian: Actually, I've changed my mind on this one. Earlier this morning we began a conversation that would allow konvoy to pick up the newly added addons and prompt the user to enable all, some, or none. This won't be part of this document.

joejulian: At this time, the expectation is that we're going to be extremely strict about the upgrade processes, requiring it to be transparent to the user. The usual semver reasons for major/minor/patch versions will not apply to this release process.

It wouldn't really make sense, anyway, since it's the underlying addons that are changing semver releases and we can't really match pace with those.

For now, I'm leaving this out of this document. We can revisit if the need arises.

joejulian: At this time, I'm going to leave the answer to this problem out of the document. Once the problem arises, we can revisit.

joejulian: I've added a very strong statement about this problem.

RELEASE.md
joejulian: Added text to make it clear that this should be using automated tooling.

PR #125, opened 2020-02-06
Add cloudproviders retroactively for traefik

shaneutt:
This retroactive (and backwards compatibly) fixes a bug in kubernetes-base-addons where traefik was not declared for all cloud providers.


alejandroEsc:
Thanks for adding these missing fields.

shaneutt:
> Did this change? If so, I'd rather fix the bug. If no providers are listed it's supposed to be enabled for all providers.

As per our discussions in zoom, since this change is innocuous can we take this conversation to https://github.com/mesosphere/konvoy/pull/1225 instead and block there instead of here? let me know what you think?

PR #140, opened 2020-02-14
Bump init-container to include the fix for static users

GoelDeepak:
**What type of PR is this?**
Bug
https://jira.d2iq.com/browse/D2IQ-64332

This PR bumps the init-container version which fixes an issue in dex addon which disallowed adding local users. The original patch is https://github.com/mesosphere/kubeaddons-extrasteps/pull/59

No

```release-note
Fixes an issue in dex addon which disallowed adding local users
```

**Checklist**


GoelDeepak:
> This needs to be a revision, and then 👍

@shaneutt I was debating it myself. The last one was created 2 days back and hasn't been released yet. So thought of updating that itself. What do you think?

PR #145, opened 2020-02-19
feat: added kiam addon for annotation role assume usage

sebbrandt87:
**What type of PR is this?**
Feature

adds the addon kiam to be used for annotation set role assuming for pods.
e.g. going to be used by velero

https://jira.d2iq.com/browse/D2IQ-64444

NONE

**Checklist**


shaneutt:
Somehow a panic in the tests fell through TeamCity and the CI passed, this has been reported as [D2IQ-64452](https://jira.d2iq.com/browse/D2IQ-64452).

shaneutt:
I've added a test group for Kiam @sebbrandt87, it should now get properly deploy tested in CI, and I will follow up on the issue in [D2IQ-64452](https://jira.d2iq.com/browse/D2IQ-64452) for future addon additions.

addons/kiam/3.x/kiam-1.yaml
```diff
@@ -0,0 +1,30 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: ClusterAddon
+metadata:
+  name: kiam
+  labels:
+    kubeaddons.mesosphere.io/name: kiam
+    kubeaddons.mesosphere.io/provides: iam
+  annotations:
+    catalog.kubeaddons.mesosphere.io/addon-revision: "3.5-1"
```
shaneutt: Addon machinery currently only supports revisions and versions that are [semver](https://semver.org/) compliant.

I see upstream that the project doesn't use semver, we will need to come up with a semver compliant scheme for this Addon in order for it to be functional.

Whatever we end up doing, we should document that in comments here so future readers understand the context.

addons/kiam/3.x/kiam-1.yaml
```diff
@@ -0,0 +1,30 @@
...
+    appversion.kubeaddons.mesosphere.io/kiam: "3.5"
```
shaneutt: we should also make this semver, for consistency.

addons/kiam/3.x.0/kiam-1.yaml
```diff
@@ -0,0 +1,30 @@
...
+    catalog.kubeaddons.mesosphere.io/addon-revision: "3.5.0-1"
+    appversion.kubeaddons.mesosphere.io/kiam: "3.5.0"
+    values.chart.helm.kubeaddons.mesosphere.io/kiam: "https://raw.githubusercontent.com/uswitch/kiam/master/helm/kiam/values.yaml"
+spec:
+  namespace: kube-system
+  kubernetes:
+    minSupportedVersion: v1.15.6
+  cloudProvider:
+    - name: aws
+      enabled: true
```
jaxesn: Do we want this enabled by default? cc @samba @jieyu

shaneutt:
@sebbrandt87 it appears the addon can not reach a healthy state when deployed, you can test this locally by running the tests in `test/`. Let @mesosphere/sig-ksphere-catalog know if you need any assistance getting this resolved.

addons/kiam/3.x.0/kiam-1.yaml
sebbrandt87: if it is working as expected a 100% yes.
this is how we should sell it!
role assuming is more secure with kiam and can be restricted to namespace etc

sebbrandt87:
> I have some concerns about whether this should be a base addon and would like to discuss this with product.

Without kiam or kube2iam funcionality we restrict several addons like velero and traefik to stay with static user credentials instead of acting with D2iQ and provide the more secure way how to grant permissions for AWS API operations (like creating a route53 entry or put something to a S3 bucket).
Please consider that, if not done already.
I see this as the most important argument to add this as a kubernetes-base-addon and at set it as an default.

joejulian:
Sorry, I should have said what the alternative would be if this does not land in base. The alternatives  would be a different kubeaddon repository. Community, or something to do with whatever the use case is for this.

My question is not so much about whether kiam should exist, but rather whether it should be part of base since this will add a feature that locks the user into a specific cloud provider and the base-addons were originally intended to provide a set of tools that would still allow a user to take the same config and apply it to a different cloud provider.

Obviously, storage and load-balancers need to change from provider to provider so this _may_ fall in to that category.

To expedite and avoid being blocked, an immediate option would be to submit this to kubeaddons-community.

joejulian:
Does this PR introduce a user-facing change?:
```
NONE
```

Is this true? No release notes or user configuration needed?

sebbrandt87:
> Does this PR introduce a user-facing change?:
> 
> ```
> NONE
> ```
> 
> Is this true? No release notes or user configuration needed?

will add more details today or tomorrow for sure, this was just the initial PR and it is still marked as WIP ;)

I am not blocked at all at the moment.

joejulian:
I only gave the project a pretty cursory look as I didn't have much time, so I'll ask questions, if that's ok.

If this is installed and utilized in a cluster and the user wants to take their cluster to Azure for cost reasons, what's going to break?

sebbrandt87:
> I only gave the project a pretty cursory look as I didn't have much time, so I'll ask questions, if that's ok.
> 
> If this is installed and utilized in a cluster and the user wants to take their cluster to Azure for cost reasons, what's going to break?

Any question is okay. Always.

As with any provider change, they would need to change their provider related credentials. In every addon where it is needed.

shaneutt:
It's been a few days since there's been updates on this one, what do we need to resolve to move forward here @joejulian / @sebbrandt87 ?

sebbrandt87:
That happens if focus changes and other tasks gets a higher priority. I want to come back to this shortly.

samba:
@sebbrandt87 At this time, we're not going to support this as a default addon. 

Please consider `kiam` as a specialized use-case, which we'd regard currently as a user workload. We can assist you in setting up an addon repo for addons that your team would like to use, that are not part of the core product.

For this to become a core addon, we'd have to explore *also* supporting similar integrations on other cloud providers.

addons/kiam/3.x.0/kiam-1.yaml
samba: At this time IMO we should *not* have this in the base addons repo, as this is very specific to a single cloud provider. When we can support similar integrations on other cloud providers, I'd reconsider.

sebbrandt87:
D2IQ-64444 #resolve done https://github.com/mesosphere/kubernetes-sre-addons

PR #150, opened 2020-02-20
nfs: Added NFS Server Provisioner addon.

Gilbert88:
**What type of PR is this?**
Feature - Support NFS Server Provisioner as a kubeaddon

This PR adds the nfsserverprovisioner addon. We need it because:

- This addon will be super useful for konvoy users who need RWX PV, especially ML users, given the fact the existing solution like AWS EFS is still inmature.

- This NFS Server Provisioner supports volume dynamic provisioning.

- The NFS server launched by the provisioner can be backed by PVC, and the provisioner can create fine grained NFS shares under the server. This is a good model and users could easily deploy NFS servers on different cloud providers (azure and google cloud), as well as on-prem.

[D2IQ-63390](https://jira.d2iq.com/browse/D2IQ-63390)

```release-note
Added the NFS Server Provisioner which supports ReadWriteMany Persistent Volume, with PV dynamic provisioning supported.
```

**Checklist**

### DIFF
--- a/addons/nfsserverprovisioner/2.3.x/nfsserverprovisioner-1.yaml
+++ b/addons/nfsserverprovisioner/2.3.x/nfsserverprovisioner-1.yaml
@@ -1,0 +1,30 @@
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: Addon
+metadata:
+  annotations:
+    appversion.kubeaddons.mesosphere.io/nfs-server-provisioner: 2.3.0
+    catalog.kubeaddons.mesosphere.io/addon-revision: 0.6.0-1
+  labels:
+    kubeaddons.mesosphere.io/name: nfsserverprovisioner
+  name: nfsserverprovisioner
+  namespace: kubeaddons
+spec:
+  kubernetes:
+    minSupportedVersion: v1.15.6
+  chartReference:
+    chart: nfs-server-provisioner
+    repo: https://mesosphere.github.io/charts/staging
+    version: 0.6.0
+    values: |
+      persistence:
+        enabled: true
+        size: 100Gi
+      storageClass:
+        name: nfsserverprovisioner
+      resources:
+        limits:
+           cpu: 1000m
+           memory: 1024Mi
+        requests:
+           cpu: 50m
+           memory: 256Mi
--- a/addons/nfsserverprovisioner/2.3.x/nfsserverprovisioner-1.yaml
+++ b/addons/nfsserverprovisioner/2.3.x/nfsserverprovisioner-1.yaml
@@ -3,7 +3,7 @@ kind: Addon
 metadata:
   annotations:
     appversion.kubeaddons.mesosphere.io/nfs-server-provisioner: 2.3.0
-    catalog.kubeaddons.mesosphere.io/addon-revision: 0.6.0-1
+    catalog.kubeaddons.mesosphere.io/addon-revision: 2.3.0-1
   labels:
     kubeaddons.mesosphere.io/name: nfsserverprovisioner
   name: nfsserverprovisioner
--- a/addons/nfsserverprovisioner/2.3.x/nfsserverprovisioner-1.yaml
+++ b/addons/nfsserverprovisioner/2.3.x/nfsserverprovisioner-1.yaml
@@ -11,6 +11,15 @@ metadata:
 spec:
   kubernetes:
     minSupportedVersion: v1.15.6
+  cloudProvider:
+    - name: aws
+      enabled: false
+    - name: azure
+      enabled: false
+    - name: docker
+      enabled: false
+    - name: none
+      enabled: false
   chartReference:
     chart: nfs-server-provisioner
     repo: https://mesosphere.github.io/charts/staging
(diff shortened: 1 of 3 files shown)


addons/nfsserverprovisioner/2.3.x/nfsserverprovisioner-1.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: Addon
+metadata:
+  annotations:
+    appversion.kubeaddons.mesosphere.io/nfs-server-provisioner: 2.3.0
+    catalog.kubeaddons.mesosphere.io/addon-revision: 0.6.0-1
```
shaneutt: This should reference the appversion, not the chart version (as some addons are helm, some are kudo we use the app version).
```suggestion
    catalog.kubeaddons.mesosphere.io/addon-revision: 2.3.0-1
```

shaneutt:
Let @mesosphere/sig-ksphere-catalog know if you need any assistance with the CI and we'll help out!

Gilbert88:
>  this should go in community. Supporting the nfs provisioner against __any__ nfs server seems like a support responsibility with a vast amount of uncertainty.

True from the perspective of support resources in the future. Basically, maintaining any NFS storage backend is not easy. I would still recommend to keep this addon as our base addon, because:

- This addon will be super useful for konvoy users who need RWX PV, especially ML users, given the fact the existing solution like AWS EFS is still inmature.
- This NFS Server Provisioner supports volume dynamic provisioning.
- The NFS server launched by the provisioner can be backed by PVC, and the provisioner can create fine grained NFS shares under the server. This is a good model and users could easily deploy NFS servers on different cloud providers (azure and google cloud), as well as on-prem.

/cc @joejulian

juliangieseke:
I tend to agree with @joejulian, base-addons is meant to be a educated set of addons that gives one a k8s base setup to work with. We should only add addons that:
- most to all of konvoy users actually use (or should use)
- we’re certain to be able to support for a while and on different k8s versions (n-2). We say things like “addons in this repo are guaranteed to be tested and working together”. If that statement is violated -> COPS Issue

I lack context to decide if this would be true for this particular addon, but in general I agree with @joejulian, thst we should put any new addons in a community (or beta/testing/similar) catslog before adding them to a repo where we committed to above statements.

addons/nfsserverprovisioner/2.3.x/nfsserverprovisioner-1.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: Addon
```
hectorj2f: As a storage addon, this should be a `ClusterAddon`

addons/nfsserverprovisioner/2.3.x/nfsserverprovisioner-1.yaml
```diff
@@ -0,0 +1,39 @@
...
+    version: 0.6.0
+    values: |
+      persistence:
+        enabled: true
+        size: 100Gi
+      storageClass:
+        name: nfsserverprovisioner
+      resources:
+        limits:
+           cpu: 1000m
```
hectorj2f: We should not set limits to the storage solution, you don't want to stop working because it needs more resources than it has.

shaneutt:
It's been some time since this had any updates, is there any help needed from @mesosphere/sig-ksphere-catalog ?

samba:
@joejulian 
> Supporting the nfs provisioner against any nfs server seems like a support responsibility with a vast amount of uncertainty.

We can, at least in principle, draw the line at supporting _the provisioner_, and not the remote NFS systems. We just have to implement a degree of tooling that lets us triage issues (in production) that verifies correct behavior of the provisioner, separate from the external NFS services. Perhaps we implement a few smoke tests to identify failures of the remote system.

@juliangieseke 
> wouldnt this addon fit in kubeaddons-enterprise?
I'd say _no_, that's not well-aligned. Enterprise is for stuff *we support*, for which customers buy our platforms. Examples would be Kafka delivered by KUDO. Even if parts of it are open source, we provide commercial support for it. Implicitly, if it's in the _enterprise_ repo, it entails support contracts.

IMO we should either _define a supportable model_ for an addon like this, or move it to a community repo (i.e. no support contract implied).

Gilbert88:
Thank you, @samba !

I would like to add some clarifications, for folks to better understand this NFS Server Provisioner model (how it works).

- This provisioner includes a built-in NFS server, and is *NOT* intended for connecting to a pre-existing NFS system/server. If customers/users have any arbitrary pre-existing NFS server/system, they should consider to use the [NFS Client Provisioner](https://github.com/kubernetes-incubator/external-storage/tree/master/nfs-client) instead, which is orthogonal to the provisioner in this PR.
- The NFS server launched by this provisioner, can be backed by PVC (like EBS) from whatever StorageClass supported in the konvoy cluster. Then, the provisioner creates NFS shares under the built-in server for dynamic volumes.

We could have kept this provisioner in our own kubeaddons repository. However, it is more of a Product decision on whether or not we want to value this provisioner to benefits users/customers who need an RWX solution (like ML users).

/cc @joejulian @juliangieseke

samba:
Thanks @Gilbert88 - in that case I need to reconsider, as this bears different commitments in terms of support liability. I will follow up.

Gilbert88:
Closing this PR, since NFS kubeaddon is hosted under a different repo
https://github.com/mesosphere/kubeaddons-kubeflow/tree/master/addons/nfsserverprovisioner/2.3.x

PR #154, opened 2020-02-25
chore: disable audit log collection in fluent-bit

GoelDeepak:
**What type of PR is this?**
Chore 

It was observed in soak cluster that audit log bloats up the number of fields in an index. We still don't know why it is happening with audit log but since the GA date is close this PR is disabling audit log collection temporarily 

https://jira.d2iq.com/browse/D2IQ-64670

### DIFF
--- a/addons/fluentbit/1.3.x/fluentbit-2.yaml
+++ b/addons/fluentbit/1.3.x/fluentbit-2.yaml
@@ -1,0 +1,82 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: Addon
+metadata:
+  name: fluentbit
+  namespace: kubeaddons
+  labels:
+    kubeaddons.mesosphere.io/name: fluentbit
+  annotations:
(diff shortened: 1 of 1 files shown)


addons/fluentbit/1.3.x/fluentbit-2.yaml
```diff
@@ -0,0 +1,82 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: Addon
+metadata:
+  name: fluentbit
+  namespace: kubeaddons
+  labels:
+    kubeaddons.mesosphere.io/name: fluentbit
+  annotations:
+    catalog.kubeaddons.mesosphere.io/addon-revision: "1.3.2-1"
```
samvantran: I think this has to be `1.3.2-2`

GoelDeepak: good catch! updated

shaneutt:
Needs signed commits @GoelDeepak and then LGTM!

PR #156, opened 2020-02-25
fix: run all tests when on Addon has been modified

makkes:
**What type of PR is this?**

Bug

Whenever a change not related to an Addon is pushed to kubernetes-base-addons, CI checks will fail (see https://github.com/mesosphere/kubernetes-base-addons/pull/155 for an example case).

But if we're bumping dependencies or changing test code we actually want to run all the tests.

https://jira.d2iq.com/browse/D2IQ-64683

```release-note
none
```

**Checklist**

### DIFF
--- a/test/scripts/test-wrapper.go
+++ b/test/scripts/test-wrapper.go
@@ -14,7 +14,6 @@ import (
 
 type groupName string
 type addonName string
-type groups map[groupName][]addonName
 
 var re = regexp.MustCompile(`^addons/([a-z]+)/?`)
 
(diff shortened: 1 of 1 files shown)


test/scripts/test-wrapper.go
```diff
@@ -14,7 +14,6 @@ import (
 
 type groupName string
 type addonName string
-type groups map[groupName][]addonName
```
shaneutt: I'm curious what the reason for removing this type alias was, it seems somewhat arbitrary? :thinking:

makkes: It is somehow arbitrary tbh. I just saw that the type was only being used once, anyway, and so I didn't see a reason for it to exist.

shaneutt: I think the idea originally was just about context and expression, but no worries changing it is not going to cause any problems.

PR #163, opened 2020-02-27
[ElasticSearch, fluentbit] Create index template

alejandroEsc:
**What type of PR is this?**

Bug

This PR is an attempt to resolve D2IQ-64809, where we inject an index template for our logs and allows us to continue injecting audit logs.

https://jira.d2iq.com/browse/D2IQ-64809

~~This is a WIP, I am attempting to coordinate the creation of the template, so changes may occur frequently until i settle into a solution.~~

```release-note
Create ElasticSearch Index Template. Require Fluentbit to deploy only after ElasticSearch deploys.
```

**Checklist**

* [X] The commit message explains the changes and why are needed.


alejandroEsc:
waiting on confirmation fom @GoelDeepak and @mesosphere/sig-ksphere-observability

samvantran:
There's a KEP, let's agree on that first before we merge :)

alejandroEsc:
> There's a KEP, let's agree on that first before we merge :)

thats the idea behind the block tag

alejandroEsc:
We are ok to review this now, the KEP has been agreed to, and we are in soak atm. Thanks again for your patience on this.

addons/elasticsearch/6.8.x/elasticsearch-2.yaml
```diff
@@ -0,0 +1,81 @@
...
+            memory: 2048Mi
+          requests:
+            cpu: 500m
+            memory: 1536Mi
+      data:
+        updateStrategy:
+          type: RollingUpdate
+        hooks:
+          drain:
+            enabled: false
```
branden: Setting `data.hooks.drain.enabled: false` removes these pre-stop and post-start hooks: https://github.com/helm/charts/blob/2fe0e135074c840b08f7517aa642b857e6a519a4/stable/elasticsearch/templates/configmap.yaml#L104-L140

The pre-stop hook looks important, it migrates data to prevent it from being lost after the node shuts down. And the post-start hook appears to add the new node to elasticsearch client config (I'm not sure how we're getting by without it). Should we add those scripts back in the hooks we're setting?

alejandroEsc: I did not set this value. And honestly I do not know who made that decision. This does not turn off the hooks, it just keeps a default set of hooks from running. I suggest that if we wish to enable that we can write it in full here while setting that to false, otherwise we will have to maintain our own chart.

branden: My mistake, I got the impression from the following comment that this PR was setting this value. If we aren't changing anything here them I'm not concerned.

alejandroEsc: Yeah, those comments arent very clear and it took me a few readings plus trials to fully grasp what the intents were. Anyway thanks for looking in and challenging this, always good to double check understanding :)

PR #164, opened 2020-02-27
[chore] remove old controller deployment hack

shaneutt:
**What type of PR is this?**
Chore

Removes an old hack and uses the new standard controller deployment method.

[D2IQ-64892](https://jira.d2iq.com/browse/D2IQ-64892)

### DIFF
--- a/addons/elasticsearch/6.8.x/elasticsearch-1.yaml
+++ b/addons/elasticsearch/6.8.x/elasticsearch-1.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: kubeaddons
   labels:
     kubeaddons.mesosphere.io/name: elasticsearch
-    # TODO: we're temporarily supporting dependency on an existing default storage class
+    # TODO - we are temporarily supporting dependency on an existing default storage class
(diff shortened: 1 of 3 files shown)


shaneutt:
Blocked waiting for CI storage solution: https://github.com/mesosphere/charts/pull/473

addons/elasticsearch/6.8.x/elasticsearch-1.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   namespace: kubeaddons
   labels:
     kubeaddons.mesosphere.io/name: elasticsearch
-    # TODO: we're temporarily supporting dependency on an existing default storage class
+    # we are temporarily supporting dependency on an existing default storage class
```
alejandroEsc: do you mean to keep the TODO?

shaneutt: yep
```suggestion
    # TODO - we are temporarily supporting dependency on an existing default storage class
```

PR #167, opened 2020-02-28
[ci] use new test.Groups

shaneutt:
**What type of PR is this?**
Chore

Uses upstream test.Groups functionality

[D2IQ-64895](https://jira.d2iq.com/browse/D2IQ-64895)


shaneutt:
The groups are working properly and also exist in kommander, enterprise but here it seems that Istio is failing.
This has been reported as https://jira.d2iq.com/browse/D2IQ-65204, but I'm going to merge this maintenance fix to make sure KBA doesn't lag behind on test framework updates while we wait for it.

PR #178, opened 2020-03-11
Update CODEOWNERS

shaneutt:
Update Prometheus codeowners.


CODEOWNERS
```diff
@@ -5,3 +5,6 @@
 # Order is important: later rules override preceding rules
 
 * @mesosphere/sig-ksphere-catalog
+
+# Prometheus
+addons/prometheus/* @gracedo
```
joejulian: These are not additive, the more specific rule replaces the global rule. So we'd prefer this:
```suggestion
addons/prometheus/* @mesosphere/sig-ksphere-catalog @gracedo
```

CODEOWNERS
```diff
@@ -5,3 +5,6 @@
 # Order is important: later rules override preceding rules
 
 * @mesosphere/sig-ksphere-catalog
+
+# Prometheus
+addons/prometheus/* @mesosphere/sig-ksphere-catalog @gracedo @branden
```
makkes:
```suggestion
addons/prometheus/* @mesosphere/sig-ksphere-observability 
```

makkes:
```suggestion
addons/prometheus/* @mesosphere/sig-ksphere-catalog @mesosphere/sig-ksphere-observability 
```

makkes: I'm fine with either one of my suggestions.

PR #185, opened 2020-03-19
Test that Prometheus is available

branden:
**What type of PR is this?**
Feature?

This adds a test that verifies Prometheus is available after its addon is deployed.

https://jira.d2iq.com/browse/D2IQ-65335

```release-note
NONE
```

**Checklist**

### DIFF
--- a/test/prometheus_test.go
+++ b/test/prometheus_test.go
@@ -1,0 +1,45 @@
+package test
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/mesosphere/kubeaddons/pkg/test"
+)
+
+func TestPrometheusGroup(t *testing.T) {
+	if err := testgroup(t, "prometheus", prometheusChecker); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func prometheusChecker(t *testing.T, cluster test.Cluster) test.Job {
+	return func(t *testing.T) error {
+		if err := kubectl("apply", "-f", "./artifacts/prometheus-checker.yaml"); err != nil {
+			return err
+		}
+
+		succeeded := false
+		timeout := time.Now().Add(time.Minute * 1)
+		for timeout.After(time.Now()) {
+			job, err := cluster.Client().BatchV1().Jobs("default").Get("prometheus-checker", metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			if job.Status.Succeeded == 1 {
+				succeeded = true
+				break
+			}
+			time.Sleep(time.Second * 1)
+		}
+
+		if !succeeded {
+			return fmt.Errorf("prometheus checker job did not succeed within timeout")
+		}
+		t.Log("prometheus checker job succeeded 🤪")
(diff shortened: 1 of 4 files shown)


test/prometheus_test.go
```diff
@@ -0,0 +1,139 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/kind/pkg/apis/config/v1alpha3"
+	"sigs.k8s.io/kind/pkg/cluster"
+
+	"github.com/mesosphere/kubeaddons/pkg/test"
+	"github.com/mesosphere/kubeaddons/pkg/test/cluster/kind"
+	"github.com/mesosphere/kubeaddons/pkg/test/loadable"
+)
+
+func TestPrometheusUp(t *testing.T) {
```
shaneutt: Having this as a `Test$_(t *testing.T)` will make it run on all CI runs, which wont be right for updates to other addons not related. We should instead make sure these tests land under the `TestPrometheusGroup()` test function.

We should model it similar to the [thanos tests](https://github.com/mesosphere/kubeaddons-kommander/blob/master/test/thanos_test.go#L12) in kommander, wherein the tests are [loaded into the test harness](https://github.com/mesosphere/kubeaddons-kommander/blob/master/test/kommander_test.go#L62) to run as part of the default (post-deployment, pre-cleanup) test plan.

It may be valid to simply rename this and remove the old one:

```suggestion
func TestPrometheusGroup(t *testing.T) {
```
However if we do that we need to make sure we handle the other members of the group the same was as was done in the previous function.

test/prometheus_test.go
```diff
@@ -0,0 +1,139 @@
...
+	if err := createNodeVolumes(3, u.String(), &node); err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		if err := cleanupNodeVolumes(3, u.String(), &node); err != nil {
+			t.Logf("error: %s", err)
+		}
+	}()
+
+	tcluster, err := kind.NewCluster(version, cluster.CreateWithV1Alpha3Config(&v1alpha3.Cluster{Nodes: []v1alpha3.Node{node}}))
```
shaneutt: We shouldn't need to deploy a new `test.Cluster` here, but instead utilize the `test.Cluster` already provisioned by the prometheus testing group UNLESS (related to above) we intend to make this the new `TestPrometheusGroup()` function, which is potentially valid.

test/prometheus_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	th.Validate()
+	th.Deploy()
+	th.Default()
+
+	close(stop)
+	wg.Wait()
+}
+
+func prometheusChecker(t *testing.T, cluster test.Cluster) test.Job {
```
shaneutt: This is the meat and potatoes of the test, and looks reasonable to me :+1:

shaneutt:
@branden just wanted to check in on this one?

branden:
I'll close this PR in favor of https://github.com/mesosphere/kubernetes-base-addons/pull/211. This PR's head branch is from my fork, which breaks CI. #211's head branch is from this fork, so it has a chance of passing CI.

PR #186, opened 2020-03-19
awsebscsiprovisioner chart version 0.3.5 -> aws-ebs-csi-driver 0.5.0

sebbrandt87:
**What type of PR is this?** 
Feature

**What this PR does/ why we need it**:
- adds the new awsebscsiprovisioner chart 0.3.5
- aws-ebs-csi-driver 0.5.0
- updated compatible CSI container images
- includes more values to be setable

**Which issue(s) this PR fixes**:
[D2IQ-64990]
[D2IQ-64992]

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

**Checklist**

* [x] The commit message explains the changes and why are needed.
* [x] The code builds and passes lint/style checks locally.
* [ ] The relevant subset of integration tests pass locally.
* [ ] The core changes are covered by tests.
* [ ] The documentation is updated where needed.

[D2IQ-64990]: https://jira.d2iq.com/browse/D2IQ-64990
[D2IQ-64992]: https://jira.d2iq.com/browse/D2IQ-64992

PR #198, opened 2020-03-26
awsebscsiprovisioner: chart bump 0.3.6

sebbrandt87:
**What type of PR is this?**
Bug

- missing `volumesnapshots/status` permission correct set for `ebs-csi-controller-sa`

[D2IQ-66019]

[D2IQ-66019]: https://jira.d2iq.com/browse/D2IQ-66019

```release-note
NONE
```

**Checklist**


addons/awsebscsiprovisioner/0.5.x/awsebscsiprovisioner-2.yaml
```diff
@@ -0,0 +1,65 @@
...
+        enabled: true
+      provisioner:
+        enableVolumeScheduling: true
+      storageclass:
+        isDefault: true
+        reclaimPolicy: Delete
+        volumeBindingMode: WaitForFirstConsumer
+        type: gp2
+        fstype: ext4
+        iopsPerGB: null
```
hectorj2f: why do we need to set this value `iopsPerGB: null` ?

addons/awsebscsiprovisioner/0.5.x/awsebscsiprovisioner-2.yaml
```diff
@@ -0,0 +1,65 @@
...
+        encrypted: false
+        kmsKeyId: null
```
hectorj2f: same as above, won't the default be already null ?

hectorj2f: It is the default value https://github.com/mesosphere/charts/blob/master/stable/awsebscsiprovisioner/values.yaml#L101, let's not scare users by setting all the default values.

addons/awsebscsiprovisioner/0.5.x/awsebscsiprovisioner-2.yaml
```diff
@@ -0,0 +1,65 @@
...
+        allowedTopologies: []
+        # - matchLabelExpressions:
```
hectorj2f: the same as the chart, the chart is a free repo why we duplicate these defaults https://github.com/mesosphere/charts/blob/master/stable/awsebscsiprovisioner/values.yaml#L104

addons/awsebscsiprovisioner/0.5.x/awsebscsiprovisioner-2.yaml
sebbrandt87: to make it visible that it is setable

addons/awsebscsiprovisioner/0.5.x/awsebscsiprovisioner-2.yaml
sebbrandt87: yes

sebbrandt87:
> Why do you want to add the default values in the chart?

to make it visible that it is setable

addons/awsebscsiprovisioner/0.5.x/awsebscsiprovisioner-2.yaml
sebbrandt87: but this is not to question here as we already have it in the revision before ;)

sreis:
> to make it visible that it is setable

@sebbrandt87 This wasn't announced, is that a new requirement for charts?

sebbrandt87:
> > to make it visible that it is setable
> 
> @sebbrandt87 This wasn't announced, is that a new requirement for charts?

https://github.com/mesosphere/kubernetes-base-addons/pull/186

hectorj2f:
@sebbrandt87 perhaps we didn't need it before in https://github.com/mesosphere/kubernetes-base-addons/pull/186. I find useless and more complicated to add all the defaults, we are losing all the abstraction that makes things simpler and to look simple. If I want to know which parameters are available, I go and check the chart.

shaneutt:
@hectorj2f @sreis @sebbrandt87 having placeholder values in addons because the maintainers like to be reminded of the available settings does not currently go against any of our contribution guidelines for addons UNLESS they somehow affect the addon operationally. Given that and given the fact that some of these historically already existed on the addon, as long as @sebbrandt87 has tested that the helm templating is "doing the right thing" and these placeholders are innocuous, I recommend deferring to the contributors and unblocking on placeholders.

samba:
Echoing @shaneutt ...

The presence of placeholder values should not be a blocking concern on this PR, specifically because the prior Addon also included them. 

We now have a new problem though: inconsistency between Addons, a few of which provide unset placeholder values, but most do not.  I'd recommend that _for now_ we remove the placeholders in favor of consistency.

Elevating certain options to visibility, in this way, is a half-measure to a broader problem. All of the addons have potentially customizable values that are _indirectly_ indicated in a metadata link. Currently this requires automation (in the case of Kommander Catalog) to fetch those values and load them into a form in a UI, for the user to customize. 

This PR and a single addon are **not** the place to tackle the visibility problem. Addressing that effectively will need to cover all addons, and an agreed strategy. The approach in this PR and its prior do not fit either of these qualifications. 

@hectorj2f If you have suggestions on how we can address the visibility problem better, I'd welcome a KEP. Please do not block this PR.

hectorj2f:
@shaneutt @samba please, proceed with whatever decision you take. 
I was expecting a better reasoning and discussion from this open PR. I prefer to stay away of discussing why showing all the placeholder values of all the addons (all its options) is or isn't the way to go. It is your call ;).

shaneutt:
@samba / @hectorj2f / @sebbrandt87 / @sreis here's a JIRA to follow up on the concerns raised here: https://jira.d2iq.com/browse/D2IQ-66946

PR #184, opened 2020-03-18
Decreased curator days to 10 and set the ES data node disk size to 50Gi

vrejabram:
**What type of PR is this?**
Bug

**What this PR does/ why we need it**:
Decreased the Curator delete_indices action days to 10 and set the ElasticSearch data node disk size to 50Gi -- https://jira.d2iq.com/browse/D2IQ-65315

**Which issue(s) this PR fixes**:
https://jira.d2iq.com/browse/D2IQ-65315

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

**Checklist**
* [x] The commit message explains the changes and why are needed.
* [x] The code builds and passes lint/style checks locally.
* [x] The relevant subset of integration tests pass locally.
* [ ] The core changes are covered by tests.
* [x] The documentation is updated where needed.

PR #203, opened 2020-03-28
elasticsearch: revert the change on pv size

jieyu:
**What type of PR is this?**

Bug

Otherwise, we'll see this error during upgrade.
    
```
[kube] 2020/03/28 01:32:26 error updating the resource
"elasticsearch-kubeaddons-data": StatefulSet.apps "elasticsearch-kubeaddons-data" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden
```

no issue. trace recent konvoy aws e2e failures, this seems to be the root case.

The bug was introduced in https://github.com/mesosphere/kubernetes-base-addons/pull/184

Konvoy e2e for this KBA branch can be tracked here:
https://github.com/mesosphere/konvoy/pull/1373

```release-note
elasticsearch: revert the PVC size to default (30G) for data nodes
```

**Checklist**

### DIFF
--- a/addons/elasticsearch/6.8.x/elasticsearch-4.yaml
+++ b/addons/elasticsearch/6.8.x/elasticsearch-4.yaml
@@ -10,7 +10,7 @@ metadata:
     # on the cluster, this hack will trigger re-queue on Addons until one exists.
     kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: "true"
   annotations:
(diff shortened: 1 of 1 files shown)


jieyu:
Yeah, one option is what @dkoshkin mentioned that there should be a way for kubeaddon to use a different config for fresh install vs. upgrade.

Another option is we increase the number of replicas to increase the capacity.

PR #222, opened 2020-04-20
chore: update PULL_REQUEST_TEMPLATE.md

armandgrillet:
**What type of PR is this?**
Chore

This is already written in `CONTRIBUTING.md` but having it in the PR template can remind people.

No issue.

```release-note
NONE
```

**Checklist**

* [X] The commit message explains the changes and why are needed.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,7 +2,10 @@
 <!-- Bug, Chore, Documentation, Feature -->
 
 **What this PR does/ why we need it**:
-<!-- Explain, without going into the details, what this PR does, and what problem it solves. -->
+<!--
+Explain, without going into the details, what this PR does, and what problem it solves.
+If you are doing an addon revision and to simplify reviews, please create a first commit in your branch only copying the previous revision to the new addon file.
```
samvantran, line 7: Is this comment about copying over the revision in the first commit necessary anymore? 

Look at the way kubeaddons-kommander updates their revisions: https://github.com/mesosphere/kubeaddons-kommander/pull/56/files

armandgrillet: Interesting, AFAIK updating a file instead of creating a new revision is not the recommended way but I might be wrong. 
@juliangieseke can you tell us if we're doing revisions in place in https://github.com/mesosphere/kubeaddons-kommander/pull/56/files because of a conscious choice or just because it's a repo that has less updates than kubernetes-base-addons?

juliangieseke: I guess both of you are kinda correct 🤷‍♂️ 
Both repos, `kubernetes-base-addons` and `kubeaddons-kommander` are versioned via selecting git branches/tags in cluster.yamls `configVersion` instead of selecting individual addon versions/revisions. I've been trying to simplify devex by updating in place and since the beforementioned is true, I dont see an issue with that approach. @shaneutt had some concerns of us accedentially breaking things tho. I didnt fully understand the reason why this approach is an issue tho. I guess if we ever want to support clients that actually treat this repo as a catalog, we might need to re-add all old versions that have been deleted before?

makkes: We'll have to carry this confusion to SIG catalog and come back after having discussed it here.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,7 +2,10 @@
 <!-- Bug, Chore, Documentation, Feature -->
 
 **What this PR does/ why we need it**:
-<!-- Explain, without going into the details, what this PR does, and what problem it solves. -->
+<!--
```
makkes, line 5: @armandgrillet I would prefer to have the hint at the very top as I did in #223. wdyt?

.github/PULL_REQUEST_TEMPLATE.md
juliangieseke: true. Also I'd love to have this clarified and documented. main reason to simplify the procedure in the first place was that it often was done wrong due to lack of docs.

.github/PULL_REQUEST_TEMPLATE.md
samvantran: ooh +1 to this suggestion

PR #231, opened 2020-04-28
test that the client can establish a pipeline connection through the service proxy

joejulian:
**What type of PR is this?**

Chore

Add a test to prove that kubectl can establish a pipeline connection (currently via spdy protocol) to the api server through the service proxy.

https://jira.d2iq.com/browse/D2IQ-65969

```release-note
None
```

**Checklist**


addons/traefik/1.7.x/tests/metallb-config.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+JP='{.status.addresses[?(.type == "InternalIP")].address}'
+NODEIP=$(kubectl get node kind-control-plane -o jsonpath="${JP}")
+SLASH24=$(sed 's@\.[0-9]\+$@@' <<< ${NODEIP})
+kubectl -n kubeaddons create configmap metallb-config --from-literal=config='{"address-pools":[{"name":"default","protocol":"layer2","addresses":["'${SLASH24}'.200-'${SLASH24}'.250"]}]}'
+
```
kensipe, line 7: wdyt about making this script pass `shellcheck`?

current output:

```
shellcheck metallb-config.sh 

In metallb-config.sh  line 5:
SLASH24=$(sed 's@\.[0-9]\+$@@' <<< ${NODEIP})
          ^-- SC2001: See if you can use ${variable//search/replace} instead.
                                   ^-------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
SLASH24=$(sed 's@\.[0-9]\+$@@' <<< "${NODEIP}")


In metallb-config.sh  line 6:
kubectl -n kubeaddons create configmap metallb-config --from-literal=config='{"address-pools":[{"name":"default","protocol":"layer2","addresses":["'${SLASH24}'.200-'${SLASH24}'.250"]}]}'
                                                                                                                                                    ^--------^ SC2086: Double quote to prevent globbing and word splitting.
                                                                                                                                                                     ^--------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
kubectl -n kubeaddons create configmap metallb-config --from-literal=config='{"address-pools":[{"name":"default","protocol":"layer2","addresses":["'"${SLASH24}"'.200-'"${SLASH24}"'.250"]}]}'

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
  https://www.shellcheck.net/wiki/SC2001 -- See if you can use ${variable//se...
```

test/manifests/local-path-storage.yaml
```diff
@@ -0,0 +1,122 @@
+# create a kubernetes-in-docker cluster and replace the standard storage class with the local-path-provisioner.
+# fsGroups are not supported in the standard storage class, so we use the rancher local-path-provisioner.
```
zmalik, line 2: in `kubeaddons-enterprise` we end up using KIND `0.8.1` which comes up with local-path-provisioner already

addons/traefik/1.7.x/tests/00-install.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kudo.dev/v1beta1
+kind: TestStep
+commands:
+  - command: docker cp kind-control-plane:/etc/kubernetes/pki/ca.crt /tmp/ca.crt
```
zmalik, line 4: these tests are assuming underlying kubernetes distro, to be kind. 
I would write the tests assuming they would be run in different distros like konvoy

joejulian:
https://jira.d2iq.com/browse/D2IQ-68296 and https://jira.d2iq.com/browse/D2IQ-68298 are follow-ups to add this test to Dispatch and to resolve the hard-coded addon references.

kuttl-test.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kudo.dev/v1alpha1
+kind: TestSuite
+manifestDirs:
+  - ./test/manifests/
+commands:
+  - command: kubectl apply -f https://mesosphere.github.io/kubeaddons/bundle.yaml
+testDirs:
+  - ./addons/traefik/1.7.x/
+startKIND: false
```
faiq, line 9: You're giving a preconfigured kind beforehand to this?

Makefile
```diff
@@ -22,3 +39,47 @@ dispatch-test: set-git-ssh
...
+
+.PHONY: delete-kind-cluster
+delete-kind-cluster:
+	kind delete cluster
+	rm $(KUBECONFIG)
+
+$(KUBECONFIG):
+	kind create cluster --wait 10s --config=test/kind/kubernetes-$(KUBERNETES_VERSION).yaml
+	mkdir -p $(ARTIFACTS)
+	kubectl kuttl test --artifacts-dir=$(ARTIFACTS) test/kind/init 2>&1 |tee /dev/fd/2 | go-junit-report -set-exit-code > dist/addons_setup_report.xml
```
faiq, line 80: This knows to use the kind cluster you made beforehand? is there anyway of specifying the path to the kubeconfig? if so, can you do it? it's going to come in handy when we finish our remote kind cluster service

kuttl-test.yaml
joejulian: Yes. See the definition of KUBECONFIG above.

Makefile
joejulian: Yes. See the definition of KUBECONFIG above.

Makefile
```diff
@@ -5,6 +5,23 @@ export ADDON_TESTS_PER_ADDON_WAIT_DURATION := 10m
 export GIT_TERMINAL_PROMPT := 1
 export ADDON_TESTS_SETUP_WAIT_DURATION := 30m
 export GOPRIVATE := github.com/mesosphere/kubeaddons
+export KUBECONFIG := kubeconfig
+
+KUTTL_VERSION=0.1.0
```
kensipe, line 10: FYI: kuttl version `0.2.0` was just released

.gitignore
```diff
@@ -0,0 +1,4 @@
+dist/
+.kube/
+bin/
+kubeconfig
```
kensipe, line 4: could we add `.DS_Store` ?

addons/traefik/1.7.x/tests/11-exec.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kudo.dev/v1beta1
+kind: TestStep
+commands:
+  - command: ./exec-test.sh
```
kensipe, line 4: I'm curious regarding the jump from index `02` to `11`

more curious regarding content of `exec-test.sh` -- what are you looking for?   perhaps it is a feature we can add to kuttl.   The testing of multiple contexts is an interesting use case.  authed, non-authed, mixed... 🤔 

The switch of server is unexpected tho... more 🤔

kuttl-test.yaml
kensipe: `startKIND: false` is the default and isn't necessary... perhaps you want it to communicate it explicitly.

It is also possible to control the internals kind start config... you could set `kindConfig=test/kind/kubernetes-1.17.5.yaml` here in the kuttl-test.yaml... (with `startKIND: true`) or start kuttl with `kuttl test --kind-config=test/kind/kubernetes-1.17.5.yaml`

addons/traefik/1.7.x/tests/11-exec.yaml
kensipe: after some thought.. I really concerned here... the switch of the kubectl context (with no return to previous state) will potentially challenge all future added tests or other tests in the testsuite.

test/kind/init/tests/00-install.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: kudo.dev/v1beta1
+kind: TestStep
+commands:
+  - command: docker cp kind-control-plane:/etc/kubernetes/pki/ca.crt /tmp/ca.crt
+  - command: docker cp kind-control-plane:/etc/kubernetes/pki/ca.key /tmp/ca.key
```
kensipe, line 5: this makes a strong assertion that this is a kind cluster and kind tests only..   this is under "kind" folder... these will not be able to used across clusters... it might be good to have comments here or a doc in the folder that sets assumptions and expectations for these tests.

test/kind/init/tests/metallb-config.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+JP='{.status.addresses[?(.type == "InternalIP")].address}'
+NODEIP=$(kubectl get node kind-control-plane -o jsonpath="${JP}")
```
kensipe, line 4: another interesting use case... may be worth considering as a feature in kuttl

test/manifests/ingress.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: networking.k8s.io/v1beta1
```
kensipe, line 1: this file isn't used (in this PR)

test/kind/init/tests/00-install.yaml
kensipe: after some more thought... I don't understand these tests... this is under "kind" folder so it isn't surprising that they are kind specific and kind only... but I don't know why we would test kind.    Is this testing konvoy components on kind?  or?

addons/traefik/1.7.x/tests/11-exec.yaml
joejulian: >I'm curious regarding the jump from index 02 to 11

Anybody that's used index-named files has come across the occasion where they had to renumber them all to insert one (sysvinit, for instance). I probably should have even left a bigger gap between 0 and 2.

> more curious regarding content of exec-test.sh -- what are you looking for? perhaps it is a feature we can add to kuttl. The testing of multiple contexts is an interesting use case. authed, non-authed, mixed... thinking

What this is for is testing that pipelining still works through the proxy. It was discovered that there was a deficiency in traefik (and possibly others) that was preventing pipeline connection upgrades (SPDY protocol in the current case).

This means that a connection from kubectl through the proxy to the api server needs to be made and needs to use one of the commands that causes the connection upgrade.

> the switch of the kubectl context (with no return to previous state) will potentially challenge all future added tests or other tests in the testsuite

Fair point. I'll change it to use it's own temporary file.

test/kind/init/tests/00-install.yaml
joejulian: This isn't really a test... It's setting up stuff. I'm just cheating and using kuttl to do it hence `kind/init`.

I originally had the yaml files in `kind/init` but they weren't getting picked up, so I moved them into "tests" so they would get picked up and executed.

test/manifests/ingress.yaml
joejulian:
```
=== RUN   kuttl
    kuttl: harness.go:339: starting setup
    kuttl: harness.go:225: running tests using configured kubeconfig.
2020/05/08 15:27:25 Ingress:default/test-ingress created
```

Makefile
joejulian: Just tested and assert is failing to match status fields. Will come back to this in a later iteration.

addons/traefik/1.7.x/tests/00-assert.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: kubeaddons.mesosphere.io/v1beta1
```
zmalik, line 1: shouldn't be merged if the catalog is going to use this as addon repository
https://jira.d2iq.com/browse/D2IQ-68485

joejulian:
This is blocked due to a catalog bug. See https://jira.d2iq.com/browse/D2IQ-68485

Makefile
kensipe: `v0.3.0` is not now... and has some useful features for this PR

addons/traefik/1.7.x/tests/00-install.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kudo.dev/v1beta1
+kind: TestStep
+commands:
+  - command: kubectl apply -f ../traefik-9.yaml
```
kensipe, line 4: `v0.3.0` feature would change to 

```
apiVersion: kudo.dev/v1beta1
kind: TestStep
apply:
  - ../traefik-9.yaml 
 ```

example: https://github.com/kudobuilder/kuttl/tree/master/pkg/test/test_data/teststep-apply

File: test/kind/init/tests/00-install.yaml [hunk 14]
```diff
@@ -0,0 +1,11 @@
...
+  - command: kubectl create namespace cert-manager
+  - command: kubectl create secret tls kubernetes-root-ca --namespace cert-manager --cert=/tmp/ca.crt --key=/tmp/ca.key
+  - command: kubectl apply -f ../../../../addons/konvoyconfig/0.0.x/konvoyconfig-1.yaml
+  - command: kubectl apply -f ../../../../addons/cert-manager/0.10.x/cert-manager-4.yaml
+  - command: ./metallb-config.sh
+  - command: kubectl apply -f ../../../../addons/metallb/0.8.x/metallb-1.yaml
```
kensipe, line 11: you may need order (meaning `metallb-config.sh` may need cert-manager-4.yaml prior to it being called or metallb-1.yaml needs to follow apply)... and if order control is needed this is a good way.

for "kubectl apply -f" that can come after commands you can now with `v0.3.0` do the following:

```
apply:
  - ../../../../addons/cert-manager/0.10.x/cert-manager-4.yaml
  - ../../../../addons/metallb/0.8.x/metallb-1.yaml

```

shaneutt:
@joejulian ping?

shaneutt:
This has been open and stale for a long time, should we close this @joejulian or should we pick it back up?

joejulian:
It's in Jira, in the backlog. Leave this alone until someone picks it up again.

PR #242, opened 2020-05-06
D2IQ-67977: valuesRemap for auth-url

jr0d:
**What type of PR is this?**

Chore

Adds valuesRemap for auth-url so these addons can be configured by kommander for SSO support

https://jira.d2iq.com/browse/D2IQ-67977

delta commit: https://github.com/mesosphere/kubernetes-base-addons/pull/242/commits/886fb6bff5ba2ba89ed1cf3ad0718a39b9bc5b01

```release-note
None
```

**Checklist**


addons/opsportal/1.1.x/opsportal.yaml
```diff
@@ -29,6 +29,8 @@ spec:
     chart: opsportal
     repo: https://mesosphere.github.io/charts/stable
     version: 0.3.12
+    valuesRemap:
```
samvantran, line 32: how come in this one you renamed the file w/o the revision number but all the other addons maintain them?

jr0d: That is how the addon exists on master: https://github.com/mesosphere/kubernetes-base-addons/tree/master/addons/opsportal/1.1.x

addons/dashboard/2.0.x/dashboard-2.yaml
```diff
@@ -31,6 +31,8 @@ spec:
     chart: kubernetes-dashboard
     repo: https://mesosphere.github.io/charts/staging
     version: 2.0.0
+    valuesRemap:
+      "ingress.annotations.traefik\\.ingress\\.kubernetes\\.io/auth-url": "ingress.auth.auth-url"
```
joejulian, line 35: Ewww... that's ugly. Can't we do this with out `\\`

jr0d: I agree, but we can't unless we change the way valuesRemap works. The helm code (what kubeaddons is using to walk maps) uses "."s to traverse nested maps. It supports escaping `.`s with a `\`. YAML also supports escaping and requires a literal `\\` to be used for key names.

jr0d: Hmm.. actually removing the `""` around the key name allows for only one escape. At least with yq.

addons/opsportal/1.1.x/opsportal.yaml
juliangieseke: I'll align opsportal with other addons after this release… initial thinking was that this repo does only support snapshots via tags anyway so history is done via git tags, but given its the only addon doing it like that its confusing and should be aligned with other addons.

addons/dashboard/2.0.x/dashboard-1.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: dashboard
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "2.0.0-beta6"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "2.0.0-1"
```
makkes, line 9: This sticks out to me, moving from beta6 to 1. Just mentioning here to make sure this is what we want.

makkes:
assigning myself as well so I can track CI checks and merge if they succeed.

jr0d:
closing per: https://github.com/mesosphere/kubernetes-base-addons/pull/315

PR #256, opened 2020-05-20
Clarify supported k8s versions in release docs

branden:
**What type of PR is this?**

Documentation

This tweaks release docs to make the supported Kubernetes versions a little more clear. It also removes a declaration about supporting the latest GA release, since 1.18 has been out for a few weeks and we don't plan to add support in the immediate future.

No issue.

```release-note
NONE
```

**Checklist**

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -20,11 +20,9 @@ Releases should be twice monthly<sup>[1](#footnote1)</sup> on the second and for
 
 ### Kubernetes
 
-The Addons in this repo are the minimum base set of supported<sup>[2](#footnote2)</sup> Addons to be installed as a suite for any supported version of Kubernetes.
+**The Kubernetes versions supported by this repo are 1.15, 1.16, and 1.17.**
 
-Kubernetes support must be maintained for the latest general release of Kubernetes and the two prior minor releases.
-At the time of this writing, the latest Kubernetes release is 1.17.2.
-Support from this repo, at this time, should cover 1.15.x, 1.16.x, and 1.17.x.
+The Addons in this repo are the minimum base set of supported<sup>[2](#footnote2)</sup> Addons to be installed as a suite for any supported version of Kubernetes.
 
 ### Branches and Tags
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -21,6 +21,7 @@ Releases should be twice monthly<sup>[1](#footnote1)</sup> on the second and for
 ### Kubernetes
 
 **The Kubernetes versions supported by this repo are 1.15, 1.16, and 1.17.**
+Support should be maintained for the latest general release of Kubernetes and the two prior minor releases.
 
 The Addons in this repo are the minimum base set of supported<sup>[2](#footnote2)</sup> Addons to be installed as a suite for any supported version of Kubernetes.
 


RELEASE.md
```diff
@@ -20,11 +20,9 @@ Releases should be twice monthly<sup>[1](#footnote1)</sup> on the second and for
 
 ### Kubernetes
 
-The Addons in this repo are the minimum base set of supported<sup>[2](#footnote2)</sup> Addons to be installed as a suite for any supported version of Kubernetes.
+**The Kubernetes versions supported by this repo are 1.15, 1.16, and 1.17.**
 
-Kubernetes support must be maintained for the latest general release of Kubernetes and the two prior minor releases.
```
joejulian, line 25: This is an important restriction that should remain documented.

RELEASE.md
```diff
@@ -20,11 +20,9 @@ Releases should be twice monthly<sup>[1](#footnote1)</sup> on the second and for
...
-At the time of this writing, the latest Kubernetes release is 1.17.2.
-Support from this repo, at this time, should cover 1.15.x, 1.16.x, and 1.17.x.
```
juliangieseke, line 27: This seems to say the same as the change above?

RELEASE.md
branden: The latest general release of Kubernetes is 1.18, so we're violating this restriction. How about I replace "must" with "should"?

RELEASE.md
branden: Yes. I'm not changing the supported versions, just restating them in a way that's more prominent.

RELEASE.md
joejulian: @samba This was a product goal from the beginning. Has this changed?

makkes: I agree that we should always support latest K8s in addition to any other version. However, afaik we're not testing the latest version, are we? So if you asked me we should change the latter and not the former. 😄

RELEASE.md
```diff
@@ -20,11 +20,9 @@ Releases should be twice monthly<sup>[1](#footnote1)</sup> on the second and for
 
 ### Kubernetes
 
-The Addons in this repo are the minimum base set of supported<sup>[2](#footnote2)</sup> Addons to be installed as a suite for any supported version of Kubernetes.
+**The Kubernetes versions supported by this repo are 1.15, 1.16, and 1.17.**
```
armandgrillet, line 23: Wouldn't we need a 1.17 release to support Kubernetes 1.17? I don't see one in the releases of this repo and I see that https://jira.d2iq.com/browse/D2IQ-68046 is integrating without seeing a related PR. Am I missing something? I ask this following a question from a customer: https://mesosphere.slack.com/archives/CUPQBG3K9/p1591107332196000

samvantran: A 1.17 release hasn't happened yet but is currently in testing (1.9.0): https://github.com/mesosphere/kubernetes-base-addons/commits/testing-1.9.0

This is what we're using to test MWT on a 1.17 cluster so the next KBA release will include a `stable-1.17-xx`

samvantran: nvm! looks like @branden had been testing 1.8.0 on a 1.17 cluster and he just released https://github.com/mesosphere/kubernetes-base-addons/releases/tag/stable-1.17-1.8.0

armandgrillet: Great, thanks for the info!

RELEASE.md
samba: At least for the moment, changing it to "should" reflects our current situation. 

We should indeed aim to keep up with Kubernetes releases, however they are _not_ guaranteed on a quarterly (3-month) basis, and occasionally shift. 

Recently added 1.17 support in konvoy 1.5, and will be adding 1.18 support soon in konvoy 1.6. I'd like us to catch up with the community when 1.19 is released, but that will depend on several elements of timing. If 1.19 is released before Konvoy 1.6 goes GA, perhaps we can support both in that timeframe -- depending on complexity of migrations.

The present assumption, based on quarterly releases, was that each Konvoy release would target the *next* GA Kubernetes release, i.e. `1.4 -> 1.16`, `1.5 -> 1.17` and so on. As our release schedule flexes, as does Kubernetes upstream, this may not be exactly tenable.  We have deviated from this pattern in previous releases (e.g. Konvoy 1.2, 1.3) but we should be mindful of it in the future, as keeping the pattern would make compatibility guarantees more intuitive.

@chrisgaun frequently reminds me that we have *no customers* asking for Kubernetes 1.18 at this time. Evidently keeping up with the community is not a _customer demand_, but tracking closely with it, especially as charts for addons and other components migrate to newer Kubernetes, will be help us minimize maintenance cost.

makkes:
@branden any update on this?

branden:
@makkes Updated :)

PR #259, opened 2020-05-21
 traefik: bump to chart 1.72.19

jieyu:
**What type of PR is this?**

Bug

The chart change improves the charts by allowing extra SSL entrypoints
being defined, and inherit SSL configurations from global SSL config. We
observed that in some customer environment, they are using advanced
configurations for their ingress, including turning on proxy protocol on
the ingress.
    
Previously, we use `extraConfigEntrypoints` to add the additional velero
minio entrypoints for the velero addon. This basically means if they
have any customization on the ingress entrypoint, we'll have to ask them
to manually specifying those configs `extraConfigEntrypoints` again for
the velero-minio entrypoint, which creates a very bad user experience.
    
This chart change improves this experiecne by introducing
`extraSSLEntrypoints` that will inherit global configurations so that
the user does not have to do anything on the additional entrypoints.
    
This patch update the Addon config accordingly to use the new
`extraSSLEntrypoints`.

https://jira.d2iq.com/browse/D2IQ-68705

```release-note
[traefik] fix the velero-minio entrypoint to inherit global ssl and proxy protocol configurations 
```

**Checklist**

### DIFF
--- a/addons/traefik/1.7.x/traefik-10.yaml
+++ b/addons/traefik/1.7.x/traefik-10.yaml
@@ -7,11 +7,11 @@ metadata:
     kubeaddons.mesosphere.io/name: traefik
     kubeaddons.mesosphere.io/provides: ingresscontroller
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "1.7.23-9"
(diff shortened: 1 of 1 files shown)


jieyu:
The AWS failure seems to be a flake due to the underlying infrastructure issue.

PR #268, opened 2020-05-28
release: stable-1.8.0

joejulian:
Release of stable-1.8.0


RELEASE_NOTES.md
```diff
@@ -1,3 +1,7 @@
+## stable-1.15-1.8.0, stable-1.16-1.8.0
+
+* \[kibana\]: Fixes an issue causing an outdated version of Kibana to be deployed to GCP. ([#249](https://github.com/mesosphere/kubernetes-base-addons/pull/249), [@branden](https://github.com/branden))
```
samvantran, line 3: 🤔 all these updated addons and yet only one release note.... looks like people been skipping the template

joejulian: Do you, as a manager, want to address that now - or should we ship what we've got?

samvantran: ship now, update after

samvantran: that said, can we open the release PR sooner (beginning of sprint) so we have more time to verify with addon owners that they in fact do not have release notes?

PR #274, opened 2020-06-02
awsebscsiprovisioner: bump to chart version 0.3.8

jieyu:
**What type of PR is this?**

Chore

This makes upgrade automatic.
    
We observed that the upgrade e2e tests on aws always fail with aws csi
driver failing to install v1beta1 volumesnapshot CRDs. This starts to
happen after we bumped the master konvoy to use 1.17 by default.
    
The issue is caused by this PR #580. On 1.17 clusters, the chart will
try to install v1beta1 CRD. On a 1.16.x cluster, v1alpha1 CRD exists.
The upgrade won't work because only v1beta1 version is listed in the CRD
definition while the storage version is v1alpha1 still.
    
The above PR states that:
 ```
atm with this an "upgrade" would fail, and there would be the need to
manual remove the v1alpha1 snapshot CRDs. Which would also purge
snapshots. This must be documented.
Previously, upgrade is manual
```
    
This patch makes this process automatic.
https://jira.d2iq.com/browse/D2IQ-68987

https://jira.d2iq.com/browse/D2IQ-68987

```release-note
awsebscsiprovisioner: the manual steps to upgrade the snapshot APIs from v1alpha1 to v1beta1 is no longer required. It has been automated in the chart CRD install hook.
```

**Checklist**

### DIFF
--- a/addons/awsebscsiprovisioner/0.5.x/awsebscsiprovisioner-4.yaml
+++ b/addons/awsebscsiprovisioner/0.5.x/awsebscsiprovisioner-4.yaml
@@ -1,0 +1,65 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: ClusterAddon
+metadata:
+  name: awsebscsiprovisioner
+  labels:
+    kubeaddons.mesosphere.io/name: awsebscsiprovisioner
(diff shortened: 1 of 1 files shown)


jieyu:
Close as a dup of https://github.com/mesosphere/kubernetes-base-addons/pull/274

PR #285, opened 2020-06-05
Fix prometheus-operator chart upgrade

branden:
**What type of PR is this?**

Bug

The prometheus-operator chart can't be upgraded past 8.10.0 due to a breaking change in the grafana subchart; see https://github.com/helm/charts/issues/21200. This PR directs the addon controller to delete the existing chart release if it can't be upgraded. Note that this does not delete the chart release's persistent volumes, so data is not lost.

No issue.

I tested this PR by deploying a cluster with Konvoy 1.5, pointing the KBA config version to this PR's branch, running `konvoy deploy addons`, and then observing that the Prometheus UI reports the updated version and Grafana displays metrics going back to the time that the cluster was originally deployed.

```release-note
NONE
```

**Checklist**

### DIFF
--- a/addons/prometheus/0.38.x/prometheus-2.yaml
+++ b/addons/prometheus/0.38.x/prometheus-2.yaml
@@ -10,7 +10,7 @@ metadata:
     # on the cluster, this hack will trigger re-queue on Addons until one exists.
     kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: "true"
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "0.38.1-1"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "0.38.1-2"
     appversion.kubeaddons.mesosphere.io/prometheus-operator: "0.38.1"
     appversion.kubeaddons.mesosphere.io/prometheus: "2.17.2"
     appversion.kubeaddons.mesosphere.io/alertmanager: "0.20.0"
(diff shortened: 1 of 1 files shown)


addons/prometheus/0.38.x/prometheus-2.yaml
```diff
@@ -22,9 +22,10 @@ metadata:
     docs.kubeaddons.mesosphere.io/grafana: "https://grafana.com/docs/"
     docs.kubeaddons.mesosphere.io/alertmanager: "https://prometheus.io/docs/alerting/alertmanager/"
     values.chart.helm.kubeaddons.mesosphere.io/prometheus: "https://raw.githubusercontent.com/mesosphere/charts/81ca01f13e88fc05bf195c95356a9cfd8902ed86/staging/prometheus-operator/values.yaml"
-    # The prometheus-operator chart from prior Konvoy releases can't be upgraded to this chart version.
-    # See https://jira.d2iq.com/browse/DCOS-62924.
-    helmv2.kubeaddons.mesosphere.io/upgrade-strategy: '[{"upgradeFrom": "<=5.19.7", "strategy": "delete"}]'
+    # The prometheus-operator chart from prior Konvoy releases can't be upgraded to 8.10.0.
+    # See https://github.com/helm/charts/issues/21200.
+    # 8.8.5 was the latest version available in mesosphere/charts before it was bumped past 8.10.0.
+    helmv2.kubeaddons.mesosphere.io/upgrade-strategy: '[{"upgradeFrom": "<=8.8.5", "strategy": "delete"}]'
```
samvantran, line 28: we've employed this strategy twice now with the prom-operator (both from major and minor upgrades). I wonder if it would make more sense to change it from `<=8.x` to `>=5.19.7`? Otherwise we'll have to keep updating this strategy version.

joejulian: The goal is to _not_ delete first. It's more efficient.

PR #289, opened 2020-06-10
release: stable-2.0.0

joejulian:
Release of stable-2.0.0


jimmidyson:
This is a 2.0.0 - I assume from the new major version that there are backwards compatibility issues. Will this affect upgrades?

joejulian:
@jimmidyson Good point. That should be noted in the release notes. I'll update that.

The answer is, yes, but only for istio. Since it's in preview, product was comfortable with that.

GoelDeepak:
Blocking it till we have final solution for istio because of the limitation in KUDO there is a decision to try helm based approach.

joejulian:
Due to the above and the limited successful soak time, closing.

PR #305, opened 2020-06-23
feat: enable traefik access logs

mhrabovcin:
**What type of PR is this?**
Feature

Enables by default traefik access logging.

no issue

The additional load that this logging would put on ELK stack was tested during the last MWT. Logging did not generate additional significant load on the logging stack.

```release-note
traefik: access log is enabled by default
```

**Checklist**

### DIFF
--- a/addons/traefik/1.7.x/traefik-15.yaml
+++ b/addons/traefik/1.7.x/traefik-15.yaml
@@ -7,7 +7,7 @@ metadata:
     kubeaddons.mesosphere.io/name: traefik
     kubeaddons.mesosphere.io/provides: ingresscontroller
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "1.7.24-1"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "1.7.24-2"
     appversion.kubeaddons.mesosphere.io/traefik: "1.7.24"
     endpoint.kubeaddons.mesosphere.io/traefik: "/ops/portal/traefik"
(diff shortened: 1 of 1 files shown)


addons/traefik/1.7.x/traefik-14.yaml
```diff
@@ -61,6 +61,8 @@ spec:
           cpu: 1000m
         requests:
           cpu: 500m
+      accessLogs:
+        enabled: false
```
makkes, line 65: doesn't this **disable** access logs?

mhrabovcin: Oh you are right, I've fixed it and changed the value to `true`.

addons/traefik/1.7.x/traefik-14.yaml
```diff
@@ -0,0 +1,135 @@
...
+spec:
+  kubernetes:
+    minSupportedVersion: v1.15.6
+  requires:
+    - matchLabels:
+        kubeaddons.mesosphere.io/name: cert-manager
+  chartReference:
+    chart: traefik
+    repo: https://mesosphere.github.io/charts/staging
+    version: 1.72.20
```
hectorj2f, line 24: could we use `1.72.21` which is the latest version ?

makkes: #304 is bumping traefik, anyway so bumping here doesn't seem necessary to me.

joejulian:
Sorry, I can't merge because the commits aren't signed. Please sign them and we should be good to merge.

PR #347, opened 2020-07-08
request enough resources

joejulian:
**What type of PR is this?**

chore

Tests were being killed, OOM.

no issue

```release-note
none
```

**Checklist**

### DIFF
--- a/Dispatchfile
+++ b/Dispatchfile
@@ -1,39 +1,54 @@
 #!starlark
 
-gitResource("src-git", url = "$(context.git.url)", revision = "$(context.git.commit)")
+gitResource("src-git", url="$(context.git.url)",
+            revision="$(context.git.commit)")
+
 
 def secretVar(name, key):
-    return k8s.corev1.EnvVarSource(secretKeyRef = k8s.corev1.SecretKeySelector(
-        localObjectReference = k8s.corev1.LocalObjectReference(name=name),
-        key = key))
+    return k8s.corev1.EnvVarSource(secretKeyRef=k8s.corev1.SecretKeySelector(
+        localObjectReference=k8s.corev1.LocalObjectReference(name=name),
+        key=key))
+
 
 dindTask("dispatch-integration-test",
-         inputs = ["src-git"],
-         steps = [
+         inputs=["src-git"],
+         steps=[
              v1.Container(
                  name="fetch-master",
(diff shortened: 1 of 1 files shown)


Dispatchfile
```diff
@@ -22,7 +22,7 @@ dindTask("dispatch-integration-test",
                  command = ["make","dispatch-test"],
                  workingDir = "/workspace/src-git",
                  env = [k8s.corev1.EnvVar(name = "DISPATCH_CI", value = "true"),
-		        k8s.corev1.EnvVar(name = "AWS_REGION", value= "us-west-2"),
+		                k8s.corev1.EnvVar(name = "AWS_REGION", value= "us-west-2"),
```
armandgrillet, line 25: I would expect this to be intended as much as the line below it, not more 🤔 . Can you confirm?

joejulian: Applied pep8 to fix the overall formatting. I think there might have been a tab in there somehow.

Dispatchfile
```diff
@@ -1,39 +1,54 @@
...
+                      k8s.corev1.EnvVar(name="AWS_SECRET_ACCESS_KEY",
+                                        valueFrom=secretVar("d2iq-dispatch-aws-creds",
+                                                            "AWS_SECRET_ACCESS_KEY")),
+                      k8s.corev1.EnvVar(name="SSH_KEY_BASE64",
+                                        valueFrom=secretVar("d2iq-dispatch-git-ssh-base64",
+                                                            "ssh-privatekey-base64"))
+                      ],
+                 resources=k8s.corev1.ResourceRequirements(
+                     requests={
+                         "cpu": k8s.resource_quantity("2000m"),
```
faiq, line 41: 2 cpus for a request to run remote tests seem quite high don't they? I would reccomend that we remove the requests and increase limits so that they can burst to the appropriate level.

joejulian: We want this to schedule to a node that has room. These requests will ensure the pod is scheduled to a node that should satisfy the needs. If we went with defaults, it would still be possible (maybe even likely) for the application to be OOM killed without reaching the limits if the node itself runs out of memory.

Since these are not permanent allocations, I'm not concerned with the size of the requests at this time. If it turns out that these pods are not getting scheduled due to there being a lack of nodes available to satisfy the requests, we could come back at this at that time.

PR #382, opened 2020-07-29
fix: fluent-bit privileged fix

sebbrandt87:
**What type of PR is this?**

Bug

- accidently set the wrong value, so /dev/kmsg 
could not have been mounted

```
jql=key in (D2IQ-70488)
```

[D2IQ-70488]

[D2IQ-70488]: https://jira.d2iq.com/browse/D2IQ-70488

```release-note
NONE
```

**Checklist**

### DIFF
--- a/addons/fluentbit/1.5.x/fluentbit-3.yaml
+++ b/addons/fluentbit/1.5.x/fluentbit-3.yaml
@@ -1,0 +1,192 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: Addon
+metadata:
+  name: fluentbit
+  namespace: kubeaddons
+  labels:
+    kubeaddons.mesosphere.io/name: fluentbit
+  annotations:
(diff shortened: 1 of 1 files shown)


sebbrandt87:
> please add a regression test for this Addon so this doesn't happen again.

We could add this after that being in. It should have failed anyway already without an extra test, as the pods were not starting at all. So maybe a general `dispatch-integration-test ` or `Konvoy - e2e (aws)` is what is needed.

drewvanstone:
> > please add a regression test for this Addon so this doesn't happen again.
> 
> @makkes
> We could add this after that being in. It should have failed anyway already without an extra test, as the pods were not starting at all. So maybe a general `dispatch-integration-test ` or `Konvoy - e2e (aws)` is what is needed.

@sebbrandt87 mind creating a JIRA to track the needed tests and link it here in the PR so we have a paper trail? Then let's get this merged in.

sebbrandt87:
@makkes @drewvanstone https://jira.d2iq.com/browse/D2IQ-70493

PR #386, opened 2020-07-29
[kube-oidc-proxy] version bump for helm3 deployment

alejandroEsc:
**What type of PR is this?**

Chore

Points to a chart bump with fixes to allow helm3 deploy the chart correctly.

https://jira.d2iq.com/browse/D2IQ-70239

```release-note
NONE
```

**Checklist**


samvantran:
Hm, should there be a release note for something like `kube-oidc-proxy now supports Helm3 deployment`? Same thing for all the other ones that were updated for this effort?

alejandroEsc:
> Hm, should there be a release note for something like `kube-oidc-proxy now supports Helm3 deployment`? Same thing for all the other ones that were updated for this effort?

Not really, a lot of the changes are not helm3 specific changes if any. They were legit chart issues that should have never made it through helm though helm seemed to be ok with them. I would say more like the changes are corrections to the helm charts themselves and not for helm3. If they were major changes like that, then yes i would have included them, in fact the only one like that would be kubefed but that most likely will be done upstream.

PR #388, opened 2020-07-30
Revert "fix: fluent-bit privileged fix"

joejulian:
Reverts mesosphere/kubernetes-base-addons#382

You were right the first time: https://github.com/fluent/helm-charts/blob/fluent-bit-0.6.1/charts/fluent-bit/templates/_pod.tpl#L11


sebbrandt87:
nope @joejulian @alejandroEsc https://github.com/fluent/helm-charts/blob/fluent-bit-0.6.1/charts/fluent-bit/templates/_pod.tpl#L15 😬 
sneaky detail 😬

PR #389, opened 2020-07-30
fix: enable serviceMonitor

sebbrandt87:
**What type of PR is this?**

Bug / Feature

- Enable serviceMonitor, disable service annotations.
- Add env for being used in filter to extend. audit.log records and 
kernel messages (the others have that data already)
- Explicitly set Refresh_Interval and Rotate_Wait for audit tail and 
containers tail.

no issue

Tested on a new created cluster.
Seen all the metrics back again.
New host key field is appended to the two configured inputs.
Configmap updated with Refresh_interval and Rotate_Wait. 🤞

```release-note
NONE
```

**Checklist**


sebbrandt87:
E2E 

```
<*errors.errorString | 0xc0004fd750>: {
s: "1 / 136 pods in namespace \"\" are NOT in RUNNING and READY state in 10m0s\nPOD                             NODE PHASE   GRACE CONDITIONS\nelasticsearch-kubeaddons-data-3      Pending       [{Type:PodScheduled Status:False LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2020-07-30 10:59:22 +0000 UTC Reason:Unschedulable Message:0/8 nodes are available: 8 Insufficient cpu.}]\n",
}
```

not related

and known

```
/root/src/github.com/mesosphere/konvoy/tests/e2e/autoscaling.go:324
error checking node pool: count should be equal or greater than 3
Unexpected error:
<*errors.errorString | 0xc00009b980>: {
s: "timed out waiting for the condition",
}
timed out waiting for the condition
occurred
/root/src/github.com/mesosphere/konvoy/tests/e2e/autoscaling.go:445
Open in build log
```

We may should disable autoscaler E2E test in kubeaddons E2E or wait longer?
For elasticsearch we would need more computing resources in the started test cluster

addons/fluentbit/1.5.x/fluentbit-4.yaml
```diff
@@ -102,19 +110,20 @@ spec:
               Parser kubernetes-audit
               DB /tail-db/audit.db
               Tag audit.*
-              Refresh_Interval 5
```
alejandroEsc: What are the pros/cons for making the interval changes here?

addons/fluentbit/1.5.x/fluentbit-4.yaml
```diff
@@ -35,10 +35,13 @@ spec:
     version: 0.6.1
     values: |
       service:
-        annotations:
-          prometheus.io/path: "/api/v1/metrics/prometheus"
-          prometheus.io/port: "2020"
-          prometheus.io/scrape: "true"
+        annotations: {}
+      serviceMonitor:
```
alejandroEsc, line 39: Could you give me a quick explanation for this change? Have you tested this out with prometheus?

sebbrandt87: Sure.
So first added annotations where for pod monitoring / scraping. These didn't worked as they we're not complete for our Prometheus operator setup.

As the official fluent-bit chart has a good Service Monitor enablement logic (with checking for api capability) I checked out we can really use it in this way. And it works for our Prometheus Operator addon.
So we we're able to drop the annotations as not used and use ServiceMonitor.

addons/fluentbit/1.5.x/fluentbit-4.yaml
sebbrandt87: Refresh and Rotation Check Interval were even before (refresh 5 and default rotation wait 5).
In bigger clusters and long running we have seen that fluent-bit loses the tailing for auditlogs.
This change is possibly fixing it and has no real cons. Except it checks less often for changed Files to tail.

addons/fluentbit/1.5.x/fluentbit-4.yaml
alejandroEsc: thank you for that explanation.

addons/fluentbit/1.5.x/fluentbit-4.yaml
```diff
@@ -35,10 +35,13 @@ spec:
     version: 0.6.1
     values: |
       service:
-        annotations:
-          prometheus.io/path: "/api/v1/metrics/prometheus"
-          prometheus.io/port: "2020"
-          prometheus.io/scrape: "true"
+        annotations: {}
```
joejulian, line 38:
```suggestion
```

addons/fluentbit/1.5.x/fluentbit-4.yaml
```diff
@@ -0,0 +1,208 @@
...
+    - name: docker
+      enabled: false
+    - name: none
+      enabled: true
+  chartReference:
+    chart: fluent-bit
+    repo: https://fluent.github.io/helm-charts
+    version: 0.6.1
+    values: |
+      service:
```
joejulian, line 37:
```suggestion
```

addons/fluentbit/1.5.x/fluentbit-4.yaml
gracedo: Just FYI, this would have had to have been a service label `servicemonitor.kubeaddons.mesosphere.io/path: "api__v1__metrics__prometheus"` and maybe a change to prom config to add the port for it to work, but it's better to use the servicemonitor defined in fluentbit since it actually checks [`.Capabilities.APIVersions.Has "monitoring.coreos.com/v1"`](url)

joejulian: > ... it actually checks .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" ...

It does? Oh no. That means that it won't install the service monitor if prometheus isn't installed first.

gracedo: Lol right, back to this recurring chart-dependency issue 🙈 Do we want to just be consistent from now on about defining **all** servicemonitors within prometheus-operator until we come up with a better solution? There are a couple addons that don't - velero and istio are two that I can think of

gracedo: Unfortunately, [fluent-bit currently doesn't support adding service labels](https://github.com/fluent/helm-charts/blob/master/charts/fluent-bit/templates/service.yaml#L5-L6) so we'd have to open up an upstream pr to add that first. In my experience (I've had to do this a couple times for upstream charts for precisely this reason - add labels for prom service monitor :p) the change is accepted without any issue. Depending on how active the chart maintainers are, it can be a quick merge (or not, if it's hard to get ahold of them)

gracedo: We'll see 🙃 https://github.com/fluent/helm-charts/pull/27

gracedo: ^ Merged! I'll bump the chart to `0.6.2`

addons/fluentbit/1.5.x/fluentbit-4.yaml
```diff
@@ -0,0 +1,207 @@
...
+    - name: gcp
+      enabled: true
+    - name: docker
+      enabled: false
+    - name: none
+      enabled: true
+  chartReference:
+    chart: fluent-bit
+    repo: https://fluent.github.io/helm-charts
+    version: 0.6.1
```
gracedo, line 35:
```suggestion
    version: 0.6.2
```

addons/fluentbit/1.5.x/fluentbit-4.yaml
gracedo: we'll also need to add an endpoint to the service monitor def in prometheus-operator [here](https://github.com/mesosphere/kubernetes-base-addons/blob/master/addons/prometheus/0.38.x/prometheus-12.yaml#L109) something like
```
              - path: /api/v1/metrics/prometheus
                port: http
                interval: 30s
```
to get metrics scraped

sebbrandt87: > It does? Oh no. That means that it won't install the service monitor if prometheus isn't installed first.

@joejulian 
What is the point?
Fluent-bit comes long after elasticsearch And that for after prometheus. We can make fluent-bit also dependent to have prometheus installed  before. We should not move ServiceMonitors out of responsible charts.


> we'll also need to add an endpoint to the service monitor def in prometheus-operator here something like
That is **not** true @gracedo

sebbrandt87: @gracedo why isn't something like
``` 
servicemonitor.kubeaddons.mesosphere.io/path: "api__v1__metrics__prometheus"
servicemonitor.kubeaddons.mesosphere.io/port: "http"
```

without any extra change to another chart or addon?

sebbrandt87: @gracedo added more labels to define it clearer also in `prometheus`(-operator) addon

sebbrandt87:
when #392 is merged, we are good to go merging, if tests are green except `continuous-integration/teamcity`
correct? @gracedo @shaneutt @alejandroEsc @joejulian

shaneutt:
Something has changed in CI, possibly related to the enabling of the service monitors, which is causing the tests to fail.

It seems our previous `elasticsearch_test.go` were not patient about readiness and would throw an error if they couldn't nearly _immediately_ port-forward.

```go
func elasticsearchChecker(t *testing.T, cluster testcluster.Cluster) testharness.Job {
	return func(t *testing.T) error {
		time.Sleep(time.Second * 120)
		localport, stop, err := portForwardPodWithPrefix(cluster, "kubeaddons", esClientPodPrefix, esClientPort)
		if err != nil {
			return fmt.Errorf("could not forward port to elasticsearch client pod: %s", err)
		}
		defer close(stop)

		if err := checkElasticsearchAvailable(localport); err != nil {
			return fmt.Errorf("failed to check elasticsearch is available: %s", err)
		}
		if err := checkElasticsearchCreateAndGetDoc(localport); err != nil {
			return fmt.Errorf("failed to create and get elasticsearch document: %s", err)
		}

		t.Logf("INFO: successfully tested elasticsearch")
		return nil
	}
}
```

I was able to get this test passing with a quick hack:

```diff
diff --git a/test/elasticsearch_test.go b/test/elasticsearch_test.go
index c3cff0a..d95c9fd 100644
--- a/test/elasticsearch_test.go
+++ b/test/elasticsearch_test.go
@@ -24,6 +24,7 @@ const (
 
 func elasticsearchChecker(t *testing.T, cluster testcluster.Cluster) testharness.Job {
        return func(t *testing.T) error {
+               time.Sleep(time.Second * 120)
                localport, stop, err := portForwardPodWithPrefix(cluster, "kubeaddons", esClientPodPrefix, esClientPort)
                if err != nil {
                        return fmt.Errorf("could not forward port to elasticsearch client pod: %s", err)
```

@mesosphere/sig-ksphere-catalog team can be made responsible for fixing this with some real readiness checking, unless you're feeling like you'd like to tackle it but we're happy to help just let us know @sebbrandt87 :vulcan_salute:

shaneutt:
As a follow up I've created [D2IQ-70677](https://jira.d2iq.com/browse/D2IQ-70677) as the long term fix for problems like these should be transient failure tolerances: this eventually succeeded so it should not have been reported as a full test failure.

sebbrandt87:
I will merge soon

PR #390, opened 2020-07-30
fix(kibana): added check if curls successful

sebbrandt87:
**What type of PR is this?**

Bug / Feature

- check for all indexPattern creations
- check for defaultIndex set
- updated plugins

exit if not successful, so that the container create-index-patterns is recreated

no issue

Tested with a new created cluster and this branch.
Indices all present and kubernetes_cluster correct set to default.

```release-note
NONE
```

**Checklist**


shaneutt:
Wasn't sure if this was `ready` or not, but it _looks_ like it does. If it is feel free to just switch the label to `ready` :vulcan_salute:

sebbrandt87:
🥈 for CI 😬 
waiting

shaneutt:
> 2nd_place_medal for CI grimacing
> waiting
> 
> unrealated autoscaler errors in CI
> 
> ```
> /root/src/github.com/mesosphere/konvoy/tests/e2e/autoscaling.go:324
> error checking node pool: count should be equal or greater than 3
> Unexpected error:
> <*errors.errorString | 0xc0000df970>: {
> s: "timed out waiting for the condition",
> }
> timed out waiting for the condition
> occurred
> /root/src/github.com/mesosphere/konvoy/tests/e2e/autoscaling.go:445
> ```

cc @dkoshkin & @dlipovetsky

sebbrandt87:
> I'm OK with merging through the Konvoy problems as long as you can please do a little ✔️ that you've tested this locally, and how it was tested. 🖖

added in PR description.

>Special notes for your reviewer:
Tested with a new created cluster and this branch.
Indices all present and kubernetes_cluster correct set to default.

addons/kibana/6.8.x/kibana-8.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: kibana
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "6.8.10-2"
```
alejandroEsc: awesome thanks for fixing this!

addons/kibana/6.8.x/kibana-8.yaml
```diff
@@ -82,11 +82,20 @@ spec:
                     -H "Content-Type: application/json; charset=utf-8" -H 'kbn-xsrf: true' \
                     -d '{"attributes":{"title":"'"${index}"'-*","timeFieldName":"@ts","fields":"[]"}}'
                 fi
+                # check or exit, to redo
+                if [[ "$(curl -s -o /dev/null -w '%{http_code}\n' -L ${KB_URL}/api/saved_objects/index-pattern/${index})" != "200" ]]; then
```
alejandroEsc, line 86: I like the patter for checking you establish on line 92 and line 95, why not convert this also?

sebbrandt87: I thought setting default at new setup and always after restart isn't a bad idea.
So we just check if the default set succeeded with checking HTTP response code after POST operation

sebbrandt87:
@alejandroEsc 
What do you think about dropping the plugin container and using default plugin Install functionality?
With the offical elasticsearch charts we would do it in the same way.
A kibana container isn't restarted often and we should use lifecycle post start functionality instead ( will show that in another addon PR 😬)

addons/kibana/6.8.x/kibana-8.yaml
sebbrandt87: confused myself 🤦 
it is done as wished 😄

addons/kibana/6.8.x/kibana-8.yaml
```diff
@@ -59,9 +59,7 @@ spec:
         requests:
           cpu: 100m
       plugins:
-        # to avoid needing to download any plugins at runtime, use a container and a shared volume
-        # do not enable the plugins here, instead rebuild the mesosphere/kibana-plugins image with the new plugins
-        enabled: false
```
alejandroEsc: So this may be something that affects us in term of airgapp. I actually do not recall the reasons for this being false, but i remember having discussions with @dkoshkin. @dkoshkin could you please give us some input here?

gracedo: Yeah this was done bc of airgap

sebbrandt87: Okay. So then we should build for now the new plugin image. Who can do that @gracedo. Is this already in a CI script?

For the next PR regarding elastic 7.8 we must build and provide hole kibana image with that plugin preinstalled.

sebbrandt87: created PR https://github.com/mesosphere/kubeaddons-sidecars/pull/2

makkes: Looking at the discussion above, I believe this should stay disabled. Plus, the comment should include a link to https://github.com/mesosphere/kubeaddons-sidecars/ so people know **how** to rebuild the image.

addons/kibana/6.8.x/kibana-8.yaml
```diff
@@ -78,30 +76,27 @@ spec:
...
+              RES_DEFAULT_INDEX=$(curl -XPOST -s -o /dev/null -w '%{http_code}\n' ${KB_URL}/api/kibana/settings/defaultIndex \
                 -H "Content-Type: application/json; charset=utf-8" -H 'kbn-xsrf: true' \
-                -d '{"value": "kubernetes_cluster"}'
+                -d '{"value": "kubernetes_cluster"}')
+              if [[ "${RES_DEFAULT_INDEX}" != "200" ]]; then
+                echo "Error within defaultIndex set - exiting"
+                exit 1
+              fi
               sleep infinity
-      initContainers:
```
makkes: This should stay as well to not break air-gapped, I suppose?!

addons/kibana/6.8.x/kibana-8.yaml
```diff
@@ -78,20 +79,28 @@ spec:
               # Create index patterns
               for index in $INDICES; do
                 if [[ "$(curl -s -o /dev/null -w '%{http_code}\n' -L ${KB_URL}/api/saved_objects/index-pattern/${index})" != "200" ]]; then
-                  curl -XPOST "${KB_URL}/api/saved_objects/index-pattern/${index}" \
+                  RES_CREATE_INDEX=$(curl -XPOST -s -o /dev/null -w '%{http_code}\n' ${KB_URL}/api/saved_objects/index-pattern/${index} \
```
makkes, line 82:
```suggestion
                  RES_CREATE_INDEX=$(curl -XPOST -s -o /dev/null -w '%{http_code}' ${KB_URL}/api/saved_objects/index-pattern/${index} \
```
the newline isn't necessary here (it doesn't hurt, anyway).

addons/kibana/6.8.x/kibana-8.yaml
```diff
@@ -78,20 +79,28 @@ spec:
...
+                    -d '{"attributes":{"title":"'"${index}"'-*","timeFieldName":"@ts","fields":"[]"}}')
+                  if [[ "${RES_CREATE_INDEX}" != "200" ]]; then
+                    echo "Error within index-pattern '${index}' creation - exiting"
+                    exit 1
+                  fi
                 fi
               done
               # Set default index pattern
-              curl -XPOST "$KB_URL/api/kibana/settings/defaultIndex" \
+              RES_DEFAULT_INDEX=$(curl -XPOST -s -o /dev/null -w '%{http_code}\n' ${KB_URL}/api/kibana/settings/defaultIndex \
```
makkes, line 92:
```suggestion
              RES_DEFAULT_INDEX=$(curl -XPOST -s -o /dev/null -w '%{http_code}' ${KB_URL}/api/kibana/settings/defaultIndex \
```
same

addons/kibana/6.8.x/kibana-8.yaml
sebbrandt87: true true, but you missed two 😉

sebbrandt87:
can we get. this in anyway continuous-integration/teamcity is broken at the moment overall elasticsearch.
I mean I tested the deployment on extra clusters created an all went well.
If you want to I can do it again just to be sure @gracedo @makkes @shaneutt

alejandroEsc:
There is nothing in this PR that should affect testing, and it seems the test fail because of some conditions are not being met and my suspicion is that its related to storage. I am ok to merge this but we need to look at this CI issue as it has affected velero also.

sebbrandt87:
> There is nothing in this PR that should affect testing, and it seems the test fail because of some conditions are not being met and my suspicion is that its related to storage. I am ok to merge this but we need to look at this CI issue as it has affected velero also.

https://jira.d2iq.com/browse/D2IQ-70658

PR #392, opened 2020-08-03
Chore: Prometheus new ServiceMonitor

sebbrandt87:
**What type of PR is this?**

Chore

We needed to add another ServiceMonitor for services with

- /api/v1/metrics/prometheus
- main port named `http`
- every 10s a scrape

related to #389 

```release-note
NONE
```

**Checklist**

### DIFF
--- a/addons/prometheus/0.38.x/prometheus-13.yaml
+++ b/addons/prometheus/0.38.x/prometheus-13.yaml
@@ -110,6 +110,19 @@ spec:
               - path: /api/v1/metrics/prometheus
                 port: metrics
                 interval: 30s
+          - name: kubeaddons-service-monitor-api-v1-metrics-prometheus-http-10s
+            selector:
+              matchLabels:
+                servicemonitor.kubeaddons.mesosphere.io/path: "api__v1__metrics__prometheus"
+                servicemonitor.kubeaddons.mesosphere.io/port: "http"
+                servicemonitor.kubeaddons.mesosphere.io/interval: "10s"
+            namespaceSelector:
+              matchNames:
+                - kubeaddons
+            endpoints:
+              - path: /api/v1/metrics/prometheus
+                port: http
+                interval: 10s
           - name: kubeaddons-service-monitor-prometheus-metrics
             selector:
               matchLabels:
--- a/addons/prometheus/0.38.x/prometheus-13.yaml
+++ b/addons/prometheus/0.38.x/prometheus-13.yaml
@@ -1,0 +1,351 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1beta2
+kind: Addon
+metadata:
+  name: prometheus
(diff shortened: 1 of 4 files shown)


addons/prometheus/0.38.x/prometheus-13.yaml
```diff
@@ -110,6 +110,19 @@ spec:
               - path: /api/v1/metrics/prometheus
                 port: metrics
                 interval: 30s
+          - name: kubeaddons-service-monitor-api-v1-metrics-prometheus-http-10s
```
gracedo, line 113: How about removing the `10s` from the def for this service monitor (including the `servicemonitor.kubeaddons.mesosphere.io/interval: "10s"` label)? in case we want to use this service monitor for other services with `api/v1/metrics/prometheus` metrics path and `http` endpoint. It also avoids us having to also modify 2 addon configs if we were to ever want to change the interval

addons/prometheus/0.38.x/prometheus-13.yaml
```diff
@@ -110,6 +110,19 @@ spec:
...
+                servicemonitor.kubeaddons.mesosphere.io/path: "api__v1__metrics__prometheus"
+                servicemonitor.kubeaddons.mesosphere.io/port: "http"
+                servicemonitor.kubeaddons.mesosphere.io/interval: "10s"
+            namespaceSelector:
+              matchNames:
+                - kubeaddons
+            endpoints:
+              - path: /api/v1/metrics/prometheus
+                port: http
+                interval: 10s
```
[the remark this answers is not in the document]
sebbrandt87: As of https://github.com/fluent/helm-charts/blob/master/charts/fluent-bit/values.yaml#L53-L59
I like to stay close. ;) 
It makes sense as fluent-bit can have heavy peaks and we would like to see that if it happens.

addons/prometheus/0.38.x/prometheus-13.yaml
sebbrandt87: We should go away from the need to manage Service Monitors in external addons not close to the addon which should be monitore. At the moment I would say we leave as is and we will see what comes next and how often we need to touch the Service Monitors here again. At best we can get rid of managing it here 😉

sebbrandt87:
Can we get this in if I approve a recheck on a new created cluster that everything is deployed and configured as it should? @gracedo @makkes @shaneutt 
`continuous-integration/teamcity` still broken as of some issues with the `ksphere-testing-framework` or `kubeaddons/utils`

makkes:
@sebbrandt87 is there a Jira tracking the failure?

sebbrandt87:
> @sebbrandt87 is there a Jira tracking the failure?

https://jira.d2iq.com/browse/D2IQ-70658

sebbrandt87:
missed to change addon revision 🤕

PR #426, opened 2020-08-06
Automated addon bumps

cprovencher:
https://jira.d2iq.com/browse/D2IQ-67176

Weekly cron job that will use "helm search" to check if newer charts are available. One pull request will be created per addon that is updated. Both chart version and annotations (addon revision, app version, docs, values) will be updated. There will be 2 commits per PR: one that copies over the latest addon yaml to a new file, and a second one that updates that new file.

Here is what the changes would look like: https://github.com/mesosphere/kubernetes-base-addons/commit/d750b947e0a351e75a1d655720fd555f5fac17e1
The commit above is only for a demo. Latest files were modified in place instead of adding a new ones to show the diff for demo purposes.


shaneutt:
Konvoy CI was a flake and TC is problematic/slow because the targeted testing script doesn't run if you don't change any addons, and so it tests the entire thing serially. The logs for that build show problems that don't seem to indicate these changes, many of the tests passed so I'm merging through to keep charles unblocked for upcoming iterations.

PR #427, opened 2020-08-07
chore: create kubernetes_audit template

sebbrandt87:
**What type of PR is this?**

Chore

- as we moved audit logs out of kubernetes_cluster index
- same curl changes as for kibana

no issue

```release-note
NONE
```

**Checklist**


sebbrandt87:
> LGTM ✔️
> 
> However, did we get a chance to manually verify? The limitations in KBA CI testing today are limited and often still leave some room for some manual testing 🤔

I tested the curl commands in that to be used container, if this is what you are asking for 😉

shaneutt:
> > LGTM heavy_check_mark
> > However, did we get a chance to manually verify? The limitations in KBA CI testing today are limited and often still leave some room for some manual testing thinking
> 
> I tested the curl commands in that to be used container, if this is what you are asking for wink

Yeah exactly, just something like that to call it "double tapped" :+1:

addons/elasticsearch/6.8.x/elasticsearch-7.yaml
```diff
@@ -0,0 +1,97 @@
...
+            # Wait until client nodes accept requests.
+            # This prevents data pods from getting into a long CrashLoopBackoff.
+            echo "Waiting for client node service ..."
+            while [[ "$(curl -s -o /dev/null -w '%{http_code}' -L ${CLIENT_URL}/_cluster/health)" != "200" ]]; do sleep 1; done
+
+            # Creating the index template: 'kubernetes_audit'
+            # Reduces the number of fields produced due to the indexing of the audit logs
+            # if template doesnt return 200, try to create it
+            if [[ "$(curl -s -o /dev/null -w '%{http_code}' -L ${CLIENT_URL}/_template/kubernetes_audit)" != "200" ]]; then
+              echo "Creating the index template: 'kubernetes_audit'"
```
alejandroEsc, line 80: I suggest we do this for kubernetes-cluster and any other indices we may need limit now or in the future. Not blocking this, but it is important to keep in mind.

addons/elasticsearch/6.8.x/elasticsearch-7.yaml
```diff
@@ -0,0 +1,97 @@
...
...ove":256}}},"name":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}},"readOnly":{"type":"boolean"}}}}},"serviceAccount":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}},"serviceAccountName":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}}}}}},"user":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}}}},"webhooks":{"properties":{"name":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}},"namespaceSelector":{"properties":{"matchExpressions":{"properties":{"key":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}},"operator":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}},"values":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}}}}}}}}}},"responseStatus":{"dynamic":false,"properties":{"code":{"type":"long"},"message":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}},"metadata":{"type":"object"},"reason":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}},"status":{"type":"text","fields":{"keyword":{"type":"keyword","ignore_above":256}}}}}}}}}')
```
branden, line 83: It shouldn't cause any problems, but I want to point out that this template includes a lot of fields that only appear on container logs, not audit logs, an artifact from when both logs were stored in the same indices. At some point we should remove these fields from the audit template and create a separate container log template that includes them.

PR #430, opened 2020-08-07
Feature: Elastic official upstream chart usage

sebbrandt87:
**What type of PR is this?**

Feature

Old helm chart is long deprecated and now it is official maintained by elastic itself.

@alejandroEsc will take care if it fits in current planned work somehow

```
jql=key in (D2IQ-70413)
```

[D2IQ-70413]

[D2IQ-70413]: https://jira.d2iq.com/browse/D2IQ-70413

```release-note
- Elastic products version 7.9.2
```

**Checklist**


samvantran:
There were some big breaking changes for elasticsearch from 6 to 7: https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html

We should be extra careful to make sure upgrades work.

Also, we might want to split up the work from switching to the official chart and the upgrade itself. That way if the upgrade causes a problem at least we can be on the official chart for 6.x.

addons/elasticsearch-data/7.8.x/elasticsearch-data-1.yaml
```diff
@@ -0,0 +1,70 @@
+---
+apiVersion: kubeaddons.mesosphere.io/v1beta1
+kind: Addon
+metadata:
+  name: elasticsearch
+  namespace: kubeaddons
+  labels:
+    kubeaddons.mesosphere.io/name: elasticsearch-data
```
samvantran, line 8: Also you're adding 3 new elasticsearch addons? Shouldn't we group all of them as subcharts? We already have a customer issue of having so many different elasticsearch addons that are actually dependent of each other and customers don't know what they can enable/disable.

shaneutt:
Note: you'll want to rebase to latest master, which includes general CI fixes.

alejandroEsc:
Let's get some motion on this chart Please. I would like the branch up-to date if possible so we can test this out. I consider this a fairly large change not to be taken lightly so some vetting is required @mesosphere/sig-ksphere-observability we probably should invest time on this soon, @samvantran Ill try to get a ticket open (if one isnt open already) to start this process.

alejandroEsc (changes requested):
I am blocking this PR after reviewing the issue a bit more. We should reconsider this approach and perhaps following this up on the multi-chart approach or the umbrella chart approach. The reason for this has to do with the chart migration process, we would need to safely remove the old addon/charts and then install these three charts. I suspect we would be ok to do some sort of upgrade strategy but that would only happen if we had one addon definition.

In addition to the the benefits of the multi-chart/umbrella approach we would be able to merge several elasticsearch charts together.

PR #437, opened 2020-08-13
[istio] update version that has helm3 fixes

alejandroEsc:
**What type of PR is this?**

Chore

Allows helm3 deployment. Nothing major is changed. 

https://jira.d2iq.com/browse/D2IQ-70994

```release-note

```

**Checklist**

### DIFF
--- a/addons/istio/1.6.x/istio-4.yaml
+++ b/addons/istio/1.6.x/istio-4.yaml
@@ -23,6 +23,7 @@ spec:
   requires:
     - matchLabels:
         kubeaddons.mesosphere.io/name: cert-manager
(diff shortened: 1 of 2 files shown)


alejandroEsc:
Ci failing due to chart not being available yet... looking as to why

alejandroEsc:
CI issue was related to prometheus being disabled. Since that has crds that istio requires, i suggest that we make this a requirement to release Istio. With that said i changed the requires field, to ensure that the order in which deployment happens is correct, and also enabled prometheus as well. Please do not disable required items from testing, if this is necessary that we must find a better solution or we are doing it wrong.

PR #449, opened 2020-08-20
Automated chart bump elasticsearch-exporter-3.7.0

d2iq-dispatch:
```release-note
Updated elasticsearch-exporter from 2.11 to 3.7.0

* Add a parameter for the elasticsearch-exporter: es.indices_settings as it is supported since version 1.0.4 (the elasticsearch-exporter chart is supporting the version 1.1.0)
* Update description for envFromSecret parameter in readme
* Feature flap the flag es.uri to allow fallback to env var ES_URI
* Allow setting environment variables with k8s secret information to support referencing already existing sensitive parameters.
* Add es.ssl.client.enabled value for better functionality readability
* Add option to disable client cert auth in Elasticsearch exporter
* Add the serviceMonitor targetLabels key as documented in the Prometheus Operator API
* Add log.level and log.format configs
* Add the ServiceMonitor metricRelabelings key as documented in the Prometheus Operator API
* Add sampleLimit configuration option
```

### DIFF
--- a/addons/elasticsearchexporter/1.1.x/elasticsearchexporter-1.yaml
+++ b/addons/elasticsearchexporter/1.1.x/elasticsearchexporter-1.yaml
@@ -1,4 +1,3 @@
----
 apiVersion: kubeaddons.mesosphere.io/v1beta1
 kind: Addon
 metadata:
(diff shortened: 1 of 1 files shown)


armandgrillet:
The release notes were done manually, right? I see `Updated elasticsearch-exporter from 2.11 to 3.0.7` right now, modifying it to be `3.7.0` before approving the PR.

armandgrillet (approved):

PR #470, opened 2020-08-25
Import bump_charts task

cprovencher:
Moved the bump_charts task to another repo (mesosphere/dispatch-tasks) so that it can be imported and used in other repos
Also changed cron to run every 2 weeks, on the 1st and 15th of every month instead of every week (that would be too many PRs)

**Checklist**


cprovencher:
> lgtm as far as i understand, just moving logic out, let's make sure we document how to find any of these bits in case something breaks.

@alejandroEsc yep, the import shows the location of the code now: https://github.com/mesosphere/kubernetes-base-addons/pull/470/files#diff-0200a01f1f9c4338c54858f69aaadae8R8

PR #485, opened 2020-09-02
Feat: Add vsphere provider

sebbrandt87:
**What type of PR is this?**

Feature

We enable and disable in addons the new `vsphere` provider as needed.

jql=key in (D2IQ-71486)

[D2IQ-71486]

[D2IQ-71486]: https://jira.d2iq.com/browse/D2IQ-71486

```release-note
- vsphere provider added
```

**Checklist**

### DIFF
--- a/addons/cert-manager/cert-manager.yaml
+++ b/addons/cert-manager/cert-manager.yaml
@@ -23,6 +23,8 @@ spec:
       enabled: true
     - name: gcp
       enabled: true
+    - name: vsphere
+      enabled: true
     - name: docker
       enabled: true
     - name: none
--- a/addons/dashboard/dashboard.yaml
+++ b/addons/dashboard/dashboard.yaml
@@ -5,7 +5,7 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: dashboard
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "2.7.1-2"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "2.7.1-3"
     appversion.kubeaddons.mesosphere.io/dashboard: "2.7.1"
     endpoint.kubeaddons.mesosphere.io/dashboard: "/ops/portal/kubernetes/"
     docs.kubeaddons.mesosphere.io/dashboard: "https://github.com/kubernetes/dashboard/blob/master/README.md"
@@ -23,6 +23,8 @@ spec:
     - name: azure
       enabled: true
     - name: gcp
+      enabled: true
+    - name: vsphere
       enabled: true
     - name: docker
       enabled: true
--- a/addons/defaultstorageclass-protection/defaultstorageclass-protection.yaml
+++ b/addons/defaultstorageclass-protection/defaultstorageclass-protection.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: defaultstorageclass-protection
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "0.0.4-2"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "0.0.4-3"
     appversion.kubeaddons.mesosphere.io/defaultstorageclass-protection: "0.0.4"
 spec:
   requires:
@@ -21,6 +21,8 @@ spec:
     - name: azure
       enabled: true
     - name: gcp
+      enabled: true
+    - name: vsphere
       enabled: true
     - name: docker
       enabled: true
--- a/addons/dex-k8s-authenticator/dex-k8s-authenticator.yaml
+++ b/addons/dex-k8s-authenticator/dex-k8s-authenticator.yaml
@@ -7,7 +7,7 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: dex-k8s-authenticator
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "1.2.2-4"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "1.2.2-5"
     appversion.kubeaddons.mesosphere.io/dex-k8s-authenticator: "v1.2.2"
     values.chart.helm.kubeaddons.mesosphere.io/dex-k8s-authenticator: "https://raw.githubusercontent.com/mesosphere/charts/f44c645c6bc843b254bb0f4f97d516f2cfee4707/staging/dex-k8s-authenticator/values.yaml"
 spec:
@@ -19,6 +19,8 @@ spec:
     - name: azure
       enabled: true
     - name: gcp
+      enabled: true
+    - name: vsphere
       enabled: true
     - name: docker
       enabled: true
--- a/addons/dex/dex.yaml
+++ b/addons/dex/dex.yaml
@@ -19,6 +19,8 @@ spec:
       enabled: true
     - name: gcp
       enabled: true
+    - name: vsphere
+      enabled: true
     - name: docker
       enabled: true
     - name: none
--- a/addons/elasticsearch-curator/elasticsearch-curator.yaml
+++ b/addons/elasticsearch-curator/elasticsearch-curator.yaml
@@ -7,7 +7,7 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: elasticsearch-curator
   annotations:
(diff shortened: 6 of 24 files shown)


shaneutt (changes requested):
@sebbrandt87 how does the default vpshere based konvoy cluster currently handle `LoadBalancer` service provisioning? Do we need to also enable [MetalLB](https://github.com/mesosphere/kubernetes-base-addons/blob/82bb5b3195c94102284c380c8f675fef0808c585/addons/metallb/0.9.x/metallb-1.yaml) as a default for this cloudProvider?

joejulian (changes requested):
We've simplified the process and you no longer have to create new files. Just update the latest of each one and increment the `addon-revision` (as you have done) see https://github.com/mesosphere/ksphere-platform/blob/master/keps/sig-ksphere-catalog/20200818-remove-revisions.md

sebbrandt87:
> Do we need to also enable MetalLB as a default for this cloudProvider?

@shaneutt yes, I missed that file to be added. Thanks for the finding! 👍 

@joejulian ❤️

addons/metallb/0.9.x/metallb-1.yaml
```diff
@@ -16,6 +16,15 @@ spec:
...
       enabled: false
+    - name: vsphere
+      enabled: true
+      values: |
+        configInline:
+          address-pools:
+          - name: default
+            protocol: layer2
+            # configure addresses for your network
+            addresses: []
```
shaneutt, line 27: This effectively means MetalLB will be broken (by default) on vsphere.

Can we provide a default IP range that will be used and available via the VSphere network deployment for our Konvoy VSphere provisioner?

shaneutt (changes requested):

addons/metallb/0.9.x/metallb-1.yaml
sebbrandt87: How should this work?
We don't know anything about customers IP ranges.
We need to document that there must be an IP range added to that list.

shaneutt: With the docker provisioner we use the default docker network range as that will be available for host networking by default.

With the VSphere provisioner is a default range for the host network not known?

sebbrandt87: I expect user setups the whole vSphere setting on its own. So also DNS, DHCP or not DHCP (so static IPs), Timeserver, etc pp.
There is basically no default.

shaneutt: Given that as the case my suggestion would be that we take this approach for now to keep current progress unblocked:

1. we remove the enabling of metallb for vpshere for now, as it can not be automatically managed as a default in current state
2. we document heavily for this iteration that manual intervention is needed to fully deploy the addons (e.g. metallb needs to be manually configured and enabled, and instructions on how to do that)
3. we create a JIRA which covers the technical need to add fully automated `LoadBalancer` svcs in the case of the VSphere provisioner
4. we make the JIRA from the above a blocker for VSphere provisioner release

Also we would need to get product team buy-in on this approach, cc @samba and @GoelDeepak.

Let me know your thoughts on this approach @sebbrandt87? :thinking:

shaneutt: (NOTE: also started a [Slack Thread](https://mesosphere.slack.com/archives/CMG6Y3C9X/p1599150474006500))

samba: @sebbrandt87 @shaneutt  I reviewed with @GoelDeepak and we've come to these changes...

* Use NodePort by default for Traefik in onprem environments & vsphere provider
* Document configuring Traefik to use LoadBalancer if available in onprem and vsphere environments
* Disable metallb by default onprem and vsphere providers
* Document the correct use of metallb to provide Service[type=LoadBalancer] in each environment
* Konvoy should look for the keepalived address and nodeport allocated to the Service for Traefik, when onprem/vsphere, instead of the current LoadBalancer external address, when printing opsportal access information

The main problems are these:
* We need sensible defaults that work out-of-the-box with zero configuration for at least a limited set of use-cases.
* We need to provide usable templates with documentation, for alternative configurations.
* We've seen numerous customer cases where third-party onprem clusters do not support Service[type=LoadBalancer], and need to accommodate this better in our addons stack.
* Konvoy will consider the cluster to be _fully functional_ only when Traefik becomes operational.

Unfortunately there's no environment where metallb can work out-of-the-box with any appropriate defaults; therefore it will require manual configuration, and therefore needs documentation. It also should therefore be disabled by default.  (I realize this is a change of direction from past ideas, but these use-cases are clarifying the needed behavior.)

We even explored the terraform vsphere provider to determine if there were good ways to pre-allocate some service IPs for metallb... and found nothing. So that's a non-starter. vSphere simply doesn't operate at the same level as other cloud providers, in respect to managed load balancers.

Naturally we'd welcome any additional feedback or alternative ideas.

samba (changes requested):
Some changes needed. This is mostly a directional reply to @shaneutt and @sebbrandt87  on the discussion of metallb and related networking concerns.

addons/metallb/0.9.x/metallb-1.yaml
shaneutt: My only feedback is that I'm generally in alignment with the proposed changes, however for this wording:

> Unfortunately there's no environment where metallb can work out-of-the-box with any appropriate defaults;

I would caveat that this is _currently_ true, but that we could look into future engineering efforts to improve our MetalLB offering in this regard which is what we started discussing in Slack.

@sebbrandt87 for your purposes to keep you moving forward: I'm glad we had this conversation but if you want to simply remove the metallb default deployment from vsphere I will consider this comment resolved and be ready to :heavy_check_mark: the PR.

shaneutt: non-blocking: just wanted to make sure I followed up with a supplemental KEP @dkoshkin and I authored which could help us improve things to better enable end-users to use MetalLB (and not have `NodePort` by default as their only option):

https://github.com/mesosphere/ksphere-platform/pull/207

shaneutt (approved):
We have a decided way forward, so I'm unblocking this as I agree with it, but just let's make sure there's no more metallb changes lingering before we merge :+1:

joejulian:
Do we have a testing strategy for this Addon?

sebbrandt87:
> Do we have a testing strategy for this Addon?

it will follow after initial alpha implementation

addons/metallb/0.9.x/metallb-1.yaml
joejulian: > > Unfortunately there's no environment where metallb can work out-of-the-box with any appropriate defaults;
> 
> I would caveat that this is currently true, but that we could look into future engineering efforts to improve our MetalLB offering in this regard which is what we started discussing in Slack.

This will always be true. There's no way to predict or interpret the desired network configuration where we don't control the network. Take an example where there are 4 network interfaces, a default on 10.1.0.0/16, a backend network on 10.2.0.0/16, a frontend network of 10.3.0.0/16, and the gateway network on 192.168.0.0/24. There's no way to interpret which network the customer wants to allocate addresses from.

shaneutt: > > > Unfortunately there's no environment where metallb can work out-of-the-box with any appropriate defaults;
> > 
> > 
> > I would caveat that this is currently true, but that we could look into future engineering efforts to improve our MetalLB offering in this regard which is what we started discussing in Slack.
> 
> This will always be true. There's no way to predict or interpret the desired network configuration where we don't control the network. Take an example where there are 4 network interfaces, a default on 10.1.0.0/16, a backend network on 10.2.0.0/16, a frontend network of 10.3.0.0/16, and the gateway network on 192.168.0.0/24. There's no way to interpret which network the customer wants to allocate addresses from.

Practically what you're saying is correct, but I don't see any argument against this: see the KEP which asks for improved flagging for the Konvoy CLI:

https://github.com/mesosphere/ksphere-platform/pull/207

sebbrandt87:
some more approves so that we can get this into master?

joejulian (approved):

GoelDeepak (approved):

sebbrandt87:
Can we get this merged now?

PR #494, opened 2020-09-04
Automated chart bump metallb-0.12.1

d2iq-dispatch:


shaneutt (approved):
This bump is OK but it's mostly ceremonial. We'll need to look into migrating to the bitnami chart soon: https://jira.d2iq.com/browse/D2IQ-71218

PR #503, opened 2020-09-04
Backport Istio 1.6.8

shaneutt:
Backport of https://github.com/mesosphere/kubernetes-base-addons/pull/489 for [D2IQ-71444](https://jira.d2iq.com/browse/D2IQ-71444)


GoelDeepak:
Why we have `cert-manager` changes here?

shaneutt:
> Why we have `cert-manager` changes here, is that intentional?

No not intentional, made a mistake with the branch targeting which should now be fixed.

GoelDeepak (approved):

samvantran:
Seems like `release-2.2` does not have `make kind-test`(and fails the tests) but `release/2.3` branch does. Should this backport go there? cc @joejulian 

is the plan to patch 2.2 for the customer or can they go to the next 2.3?

shaneutt:
> Seems like `release-2.2` does not have `make kind-test`(and fails the tests) but `release/2.3` branch does. Should this backport go there? cc @joejulian
> 
> is the plan to patch 2.2 for the customer or can they go to the next 2.3?

I talked to @joejulian about this yesterday and he pointed me to 2.3 for future Konvoy 1.5.x releases, but maybe I should just put it in both branches?

joejulian:
Unless there's a specific creq that requires this be shipped outside the normal release cycle, this should not be backported at all, and should just land in master for soak testing.

shaneutt:
> Unless there's a specific creq that requires this be shipped outside the normal release cycle, this should not be backported at all, and should just land in master for soak testing.

This is a security update for our end-users previewing Istio and includes a large amount of CVE fixes, the security aspect of this would seem to justify the backport from my perspective.

shaneutt:
> Seems like `release-2.2` does not have `make kind-test`(and fails the tests) but `release/2.3` branch does. Should this backport go there? cc @joejulian
> 
> is the plan to patch 2.2 for the customer or can they go to the next 2.3?

I have patched the testing code so that it can pass, but I'm skipping the `kind-test` path for backports, because it would bring along changes we don't yet want to backport to older Konvoy versions (relevant comment added in the `Makefile`).

shaneutt:
@joejulian you didn't block, but just wanted to check if you wanted us to hold on merging based on your previous comments?

shaneutt:
talked about this in standup, we decided we don't need this backport.

PR #511, opened 2020-09-12
Fix kibana dashboard import

branden:
**What type of PR is this?**

Bug

This fixes the dashboard import script for Kibana by overriding a couple default chart values. It sets `dashboardImport.basePath` to an empty string to prevent the dashboard import script from making requests to locations that return 404. It also increases the timeout from 60 to 180 seconds to give the Kibana API more time to become ready before the import script executes.

The dashboard import script's API queries are returning 404 because it requests paths with an extra beginning slash, e.g. `//api/status`. This worked until we upgraded Kibana from 6.8.2 to 6.8.10. Kibana 6.8.7 upgrades nodejs to 10.19.0, which introduces stricter HTTP parsing, which may explain the change. The Kibana chart deploys Kibana 6.7.0 by default, which is likely why the upstream bug hasn't been fixed yet.

https://jira.d2iq.com/browse/D2IQ-71039

```release-note
Fixes an issue that causes Kibana to deploy without an audit log dashboard.
```

**Checklist**

* [ ] The core changes are covered by tests. (A corresponding test will arrive in an upcoming PR, see https://jira.d2iq.com/browse/D2IQ-71039).
* [ ] Manually tested upgrades from previous addon revisions for updated addons.


samvantran (approved):

joejulian (approved):

alejandroEsc (approved):
LGTM

PR #513, opened 2020-09-15
[D2IQ-68980] Update to v0.19.x libs and Kubeaddons v0.20.x

shaneutt:
**What type of PR is this?**
Chore

D2IQ-71740

NONE

**Checklist**

* [x] Manually tested upgrades from previous addon revisions for updated addons.


test/go.mod
```diff
@@ -2,39 +2,22 @@ module github.com/mesosphere/kubeaddons-base-addons/test
 
 go 1.15
 
-replace k8s.io/client-go => k8s.io/client-go v0.19.0
+replace (
+	github.com/docker/docker => github.com/docker/docker v1.4.2-0.20200203170920-46ec8731fbce // locked to this version to avoid upgrades from kind that would drop our volume provisioning code. TODO: we should update/remove our volume provisioning code.
+	google.golang.org/grpc => google.golang.org/grpc v1.26.0 // held back to accomodate controller-runtime which has not updated for v1.3x.x+ at the time of writing
```
makkes, line 7: I'm not sure I understand why this needs to be replaced. What is the dependency causing this problem? I can't see any requirement of grpc in [controller-runtime's go.mod](https://github.com/kubernetes-sigs/controller-runtime/blob/master/go.mod). I'd mainly like to understand and clarify on this replacement.

test/go.mod
```diff
@@ -2,39 +2,22 @@ module github.com/mesosphere/kubeaddons-base-addons/test
...
+	k8s.io/client-go => k8s.io/client-go v0.19.1 // locked for general sanity in k8s lib resolution
```
makkes, line 8: I would prefer to not replace any version if it's not strictly necessary. Is anything breaking without this directive?

test/go.mod
shaneutt: I may have mistaken which module was doing this, I'll review one second.

test/go.mod
shaneutt: Yes, if we don't pin this it will resolve to a more recent REF which may be broken, we need to pin this one.

test/go.mod
shaneutt: This was a mistake, I worked around this already and this can be removed now :+1:

makkes (approved):

armandgrillet (approved):

PR #517, opened 2020-09-17
More test logging, and update for Kubeaddons v0.20.3

shaneutt:
**What type of PR is this?**
Chore

Adds more logging for CI tests, and updates to latest `v0.20.3` Kubeaddons which includes fixes for test utils to ensure upgrade tests run properly.

**Checklist**

### DIFF
--- a/test/addons_test.go
+++ b/test/addons_test.go
@@ -55,21 +55,25 @@ type clusterTestJob func(*testing.T, testcluster.Cluster) testharness.Job
 func init() {
 	var err error
 
+	fmt.Println("initializing local repository for test...")
 	localRepo, err = local.NewRepository("local", "../addons")
 	if err != nil {
 		panic(err)
 	}
 
+	fmt.Printf("initializing remote repository %s for test...\n", comRepoURL)
 	comRepo, err = git.NewRemoteRepository(comRepoURL, comRepoRef, comRepoRemote)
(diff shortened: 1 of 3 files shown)


joejulian (approved):

test/addons_test.go
```diff
@@ -281,14 +286,18 @@ func testgroup(t *testing.T, groupname string, version string, jobs ...clusterTe
 		}
 
 		t.Logf("determining old and new versions for upgrade testing addon %s", newAddon.GetName())
-		oldVersion, err := semver.Parse(strings.TrimPrefix(oldAddon.GetAnnotations()[constants.AddonRevisionAnnotation], "v"))
```
alejandroEsc, line 284: Do we have any protection or failure in the case where we do not have this sent to use, i.e., revision is not included?

alejandroEsc (approved):
LGTM, one question but non-blocking.

test/addons_test.go
shaneutt: Yes, the tests will fail because the `semver.Parse()` will fail.

alejandroEsc: Could we upgrade the revision annotation to a mandatory field soon?

shaneutt: We should.

PR #583, opened 2020-10-09
Automated chart bump elasticsearch-1.32.5

d2iq-dispatch:


branden:
The last time we bumped the ES chart to this version it broke up upgrades and we had to revert. See #512.

branden:
Since this ES chart is deprecated, it may make more sense to bite the bullet and move to the new upstream chart. https://jira.d2iq.com/browse/D2IQ-72433

PR #589, opened 2020-10-09
fix: fluentbit should not keep retrying indefinitely

alejandroEsc:
**What type of PR is this?**

BUG

Keep retrying on errors causes us to keep them perpetually in memory and fill out our buffer choking off fluentbit from pushing logs

```release-note
Fix fluentbit configuration to unblock output buffer. 
```

**Checklist**

### DIFF
--- a/addons/fluentbit/fluentbit.yaml
+++ b/addons/fluentbit/fluentbit.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: fluentbit
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "1.5.6-2"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "1.5.6-3"
     appversion.kubeaddons.mesosphere.io/fluentbit: "1.5.6"
     values.chart.helm.kubeaddons.mesosphere.io/fluentbit: "https://raw.githubusercontent.com/fluent/helm-charts/5e20fbc/charts/fluent-bit/values.yaml"
     # the older versions were being deployed from stable/fluent-bit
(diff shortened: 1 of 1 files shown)


gracedo (commented):
Can you bump the addon-revision, otherwise lgtm!

alejandroEsc:
> Can you bump the addon-revision, otherwise lgtm!

duh, sorry

gracedo (approved):

addons/fluentbit/fluentbit.yaml
```diff
@@ -177,7 +177,7 @@ spec:
               Time_Key @ts
               Logstash_Format On
               Logstash_Prefix kubernetes_audit
-              Retry_Limit False
+              Retry_Limit 10
```
branden, line 180: I think we'll need to set this to a low number. The backoff is pretty aggressive, chunks that will never succeed can quickly get into several minutes between retries, and there are enough of them that this can easily cause audit log collection to stop. I found that setting this to 1 on soak worked well. Let's also add a comment explaining why we need to drop audit logs more aggressively.

```suggestion
              # A small fraction of audit logs will be rejected by Elasticsearch because values may have an inconsistent type.
              # We set a low retry limit here in order to prevent fluent-bit from backing up while it repeatedly retries flushing these logs.
              # See https://github.com/uken/fluent-plugin-elasticsearch#random-400---rejected-by-elasticsearch-is-occured-why.
              Retry_Limit 1
```

branden (changes requested):

branden (approved):

branden:
dispatch-e2e-test is failing because Kommander is currently broken. I'll go ahead and merge.

PR #605, opened 2020-10-20
kibana: Update chart location, upgrade version

gracedo:
**What type of PR is this?**

Chore

Update kibana repo to elastic and update kibana to 6.8.12

https://jira.d2iq.com/browse/D2IQ-72459

```release-note
kibana: Update to 6.8.12
```

**Checklist**

### DIFF
--- a/addons/kibana/kibana.yaml
+++ b/addons/kibana/kibana.yaml
@@ -36,8 +36,6 @@ spec:
       "ingress.annotations.traefik\\.ingress\\.kubernetes\\.io/auth-url": "ingress.auth.auth-url"
     values: |
       ---
-      image:
-        tag: 6.8.10
       files:
         kibana.yml:
           ## Default Kibana configuration from kibana-docker.
--- a/addons/kibana/kibana.yaml
+++ b/addons/kibana/kibana.yaml
@@ -29,7 +29,7 @@ spec:
     - matchLabels:
         kubeaddons.mesosphere.io/name: elasticsearch
   chartReference:
-    chart: elastic/kibana
+    chart: kibana
     repo: https://helm.elastic.co
     version: 6.8.12
     valuesRemap:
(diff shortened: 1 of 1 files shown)


addons/kibana/kibana.yaml
```diff
@@ -29,14 +29,13 @@ spec:
...
-    version: 3.2.7
+    chart: elastic/kibana
+    repo: https://helm.elastic.co
+    version: 6.8.12
     valuesRemap:
       "ingress.annotations.traefik\\.ingress\\.kubernetes\\.io/auth-url": "ingress.auth.auth-url"
     values: |
       ---
-      image:
-        tag: 6.8.10
```
gracedo, line 39: @alejandroEsc @branden I wasn't sure why the image was pinned to 6.8.10 here, but if there is a particular reason why I can revert

makkes (approved):

addons/kibana/kibana.yaml
alejandroEsc: could you check the default version? I am not sure about this either, odd.

addons/kibana/kibana.yaml
```diff
@@ -29,14 +29,13 @@ spec:
     - matchLabels:
         kubeaddons.mesosphere.io/name: elasticsearch
   chartReference:
-    chart: stable/kibana
-    version: 3.2.7
+    chart: elastic/kibana
```
alejandroEsc, line 32:
```suggestion
    chart: kibana
```

alejandroEsc: the reason for this is because of the way kubeaddons defines the name of a repo (repoName) and how the chart's name is used. Otherwise CI will fail.

alejandroEsc (approved):
One item, lgtm, thank you!

addons/kibana/kibana.yaml
gracedo: The default (for 6.8.12 chart) is, unsurprisingly, 6.8.12 :p same as the previous version 6.8.10

gracedo:
Turns out the elastic kibana chart is configured completely differently than the helm chart, so we won't be able to just switch chart repos like this. Closing

PR #615, opened 2020-10-23
Automated chart bump elasticsearch-1.32.5

d2iq-dispatch:
Add release note or "none":
```release-note
```


samvantran:
Again, this is the broken elastic chart that we reverted in the past. Please do not merge any bumps to elasticsearch-1.32.5 see #512

PR #617, opened 2020-10-23
external-dns: Add servicemonitor metrics label

gracedo:
**What type of PR is this?**

Feature

Enables metrics collection by adding the servicemonitor label

https://jira.d2iq.com/browse/D2IQ-72698

```release-note
external-dns: Add servicemonitor label to enable metrics collection by Prometheus
```

**Checklist**

### DIFF
--- a/addons/external-dns/external-dns.yaml
+++ b/addons/external-dns/external-dns.yaml
@@ -6,9 +6,9 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: external-dns
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "0.7.0-2"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "0.7.0-3"
     appversion.kubeaddons.mesosphere.io/external-dns: "0.7.0"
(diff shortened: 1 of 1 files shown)


gracedo:
Tested manually on a cluster
image
image

addons/external-dns/external-dns.yaml
```diff
@@ -24,9 +24,12 @@ spec:
...
+    repo: https://mesosphere.github.io/charts/stable
     chart: external-dns
-    version: 2.20.4
+    version: 2.20.5
     values: |
       rbac:
         create: true
+      service:
+        labels:
+          servicemonitor.kubeaddons.mesosphere.io/path: "metrics"
```
samvantran, line 35: you don't need to add a name like for other fixes? eg [like this](https://github.com/mesosphere/charts/pull/851/files#diff-e1d2b0bcd42907e6b6290c1d289d57454e245189bb550886e6746ea7a38f4944R9)

samvantran (approved):
had a question but not blocking since your manual tests show it works 👍

branden (approved):

addons/external-dns/external-dns.yaml
gracedo: for external-dns, we can reuse an existing servicemonitor that just selects on that one label (https://github.com/mesosphere/kubernetes-base-addons/blob/master/addons/prometheus/prometheus.yaml#L83-L86)

PR #626, opened 2020-10-27
Add Prometheus service monitor for flagger-loadtester

branden:
**What type of PR is this?**

Bug

This adds a service monitor and related service labels so that Prometheus can discover flagger-loadtester's metrics endpoint.

I tested this PR by deploying KBA from this branch, browsing to `/ops/portal/prometheus/targets`, and observing that the flagger-loadtester target is present and available:

Screen Shot 2020-10-30 at 12 10 54 PM

https://jira.d2iq.com/browse/D2IQ-72694

```release-note
Restores metrics collection for flagger-loadtester.
```

**Checklist**

### DIFF
--- a/addons/flagger/flagger.yaml
+++ b/addons/flagger/flagger.yaml
@@ -8,9 +8,9 @@ metadata:
   annotations:
     sidecar.istio.io/inject: "false"
     appversion.kubeaddons.mesosphere.io/flagger: "0.19.0"
-    catalog.kubeaddons.mesosphere.io/addon-revision: "0.19.0-1"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "0.19.0-2"
     docs.kubeaddons.mesosphere.io/flagger: "https://docs.flagger.app/"
-    values.chart.helm.kubeaddons.mesosphere.io/flagger: "https://raw.githubusercontent.com/weaveworks/flagger/c6f3a87/charts/flagger/values.yaml"
+    values.chart.helm.kubeaddons.mesosphere.io/flagger: "https://raw.githubusercontent.com/mesosphere/charts/f4105ebb01fc758a4af356069a8ceae043201057/staging/flagger/values.yaml"
 spec:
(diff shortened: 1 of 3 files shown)


gracedo (approved):
Would you be able to verify this manually on a cluster that these metrics are getting scraped into prom?

branden:
Looks like https://github.com/mesosphere/charts/pull/860 broke the flagger chart when I changed the CRD resource version without updating the spec to match the new schema. I'll need to fix that before this PR can proceed.

addons/flagger/flagger.yaml
```diff
@@ -8,9 +8,9 @@ metadata:
   annotations:
     sidecar.istio.io/inject: "false"
     appversion.kubeaddons.mesosphere.io/flagger: "0.19.0"
-    catalog.kubeaddons.mesosphere.io/addon-revision: "0.19.0-1"
```
armandgrillet, line 11: If we bump the chartReference.version to `0.20.1`, should we update this and the `appversion.kubeaddons.mesosphere.io/flagger` to `0.20` too?

branden: I think it was a coincidence that the app and chart version were the same. Chart version 0.20.1 still deploys flagger 0.19.0: https://github.com/mesosphere/charts/blob/659d0ddada7a74e57e6a73f5a0de01fcf633b151/staging/flagger/Chart.yaml#L3-L4

s12chung: I agree with @branden. `appversion` is the internal app version inside of the chart.

s12chung (approved):

branden:
The teamcity job is failing because of a bug in the chart that's already present: https://jira.d2iq.com/browse/D2IQ-73316 This bug occurs when flagger is deployed, removed, and deployed again. This PR is triggering it because it's the first change to flagger since we added a set of tests that causes flagger to be deployed a second time to the test cluster.

PR #630, opened 2020-10-30
Automated chart bump kubernetes-dashboard-2.8.2

d2iq-dispatch:
Add release note or "none":
```release-note
```


shaneutt (changes requested):
Why is this trying to bump two different addons, where another open PR is already covering Ambassador?

PR #664, opened 2020-11-12
add requires cert-manager: v1

s12chung:
**What type of PR is this?**
Chore

Ensure that cert-manager upgrade works, by ensuring that adding wait for cert-manager v1

D2IQ-72804

```release-note
NONE
```

**Checklist**

### DIFF
--- a/addons/dex/dex.yaml
+++ b/addons/dex/dex.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     kubeaddons.mesosphere.io/name: dex
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "2.22.0-11"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "2.22.0-12"
     appversion.kubeaddons.mesosphere.io/dex: "2.22.0"
     values.chart.helm.kubeaddons.mesosphere.io/dex: "https://raw.githubusercontent.com/mesosphere/charts/5862483/stable/dex/values.yaml"
 spec:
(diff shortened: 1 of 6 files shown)


gracedo (approved):
hmm does dex need dependency on cert-manager? we had to make the cert-manager api updates in the chart but the addon doesn't have the dependency.. should it 🤔  (I'm cross-referencing this jira https://jira.d2iq.com/browse/D2IQ-70659 which included all the addons we needed to update with cert-manager v1)

s12chung:
pinged @jr0d

jr0d:
@s12chung lgtm, but I think if we want to be paranoid, any addon that requires an ingresscontroller, dex, dex-k8s-authenticater, traefik-forward-auth, kube-oidc-proxy, etc also rely on cert-manager. I think we can hold off on this for now, but keep it staged in case we run into the dependency again.

addons/dex/dex.yaml
```diff
@@ -26,6 +26,9 @@ spec:
   requires:
     - matchLabels:
         kubeaddons.mesosphere.io/provides: ingresscontroller
+    - matchLabels:
```
gracedo, line 29: dupe
```suggestion
```

s12chung: I believe we need this to match different addons:

https://github.com/mesosphere/kubernetes-base-addons/blob/f7d5de0f1be7249cc8f1f02c94047f14ae5ecf44/addons/gcpdiskprovisioner/gcpdiskprovisioner.yaml#L14-L18

joejulian: No, not a dup
There are two separate dependencies being matched.

joejulian (approved):

PR #706, opened 2020-11-30
wip: initial stab at reloader e2e test

makkes:
refs https://jira.d2iq.com/browse/D2IQ-72909

**What type of PR is this?**

```release-note

```

**Checklist**


newtest/reloader_test.go
```diff
@@ -0,0 +1,255 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/rest"
+)
+
+func deployKubeaddonsController(cluster cluster.Cluster) error {
```
shaneutt, line 22: Out of curiosity (since I tried to anticipate this need) did you not see, or not like the available helper function https://github.com/mesosphere/dkp-test-framework/blob/main/pkg/addons/kubeaddons.go#L51?

newtest/reloader_test.go
```diff
@@ -0,0 +1,255 @@
...
+	return addons.HelmInstall(cluster, "kubeaddons", "https://mesosphere.github.io/kubeaddons/chart", "kubeaddons", "kube-system", nil)
+}
+
+func deployReloader(restConfig rest.Config) error {
+	restConfig.ContentConfig.GroupVersion = &schema.GroupVersion{Group: "kubeaddons.mesosphere.io", Version: "v1beta1"}
+	client, err := rest.RESTClientFor(&restConfig)
```
shaneutt, line 28: Was https://github.com/mesosphere/dkp-test-framework/blob/main/pkg/addons/kubeaddons.go#L21 not able to replace the need for this function? Can we improve it in some way?

newtest/reloader_test.go
```diff
@@ -0,0 +1,255 @@
...
+		Data: map[string]string{
+			"index.html": `original`,
+		},
+	}
+	_, err := client.CoreV1().ConfigMaps("default").Create(context.Background(), &cm, metav1.CreateOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("could not create nginx ConfigMap: %w", err)
+	}
+
+	deployment := appsv1.Deployment{
```
shaneutt, line 58: just a nit and feel free to ignore but you could place this as a global var at the bottom as a `defaultNginxDeployment` or something like that to reduce clutter in the function.

shaneutt (commented):
I know this is WIP, these are just a couple early comments.

newtest/reloader_test.go
makkes: That's part of my feedback. Honestly, I don't believe kubeaddons should be forced onto users of this framework. We might possibly enable it through an additional Go Module but kubeaddons hauls in so many deps and requires consumers to `replace` deps that I'd rather not like to include it anywhere in my code.

newtest/reloader_test.go
makkes: see my other comment.

newtest/reloader_test.go
makkes: I'm not such a big fan of globals. ;-)

newtest/reloader_test.go
shaneutt: Not feeling like we're able to use libraries that are intended to be used as libraries seems like a huge fundamental problem to me. I'll think on this for a bit :thinking:

makkes: I'm also not sure from a scope level of dkp-test-framework whether it should expose anything kubeaddons at all. That's something we should discuss.

jimmidyson: Kubeaddons does bring with it a lot of dependencies, which of course if a good thing to avoid if possible. One thing that we did with konvoy and yakcl is to have a separate API (and indeed client) module that can be imported with minimal dependencies. This might be a good idea for kubeaddons to reduce the amount of transitive dependencies being imported into consumers of this test framework. Thoughts?

newtest/reloader_test.go
jimmidyson: Generally, I agree with the no globals rule.

newtest/reloader_test.go
makkes: I agree with that @jimmidyson said above and that's actually what I meant to say when I said
> We might possibly enable it through an additional Go Module

shaneutt: I'm in favor of splitting up the kubeaddons lib into multiple go modules for the goal of being able to use it in the testing framework, as I feel it can be invaluable for test cases.

I've created https://jira.d2iq.com/browse/D2IQ-73374 as a follow up.

In the meantime, would you be amenable to switching to the functionality that exists for the time being? Or what are your thoughts on that?

newtest/reloader_test.go
```diff
@@ -0,0 +1,255 @@
...
+		return nil, fmt.Errorf("could not create nginx Service: %w", err)
+	}
+
+	return &cm, nil
+}
+
+func curl(c typedcorev1.PodInterface, url string) (string, error) {
+	pod, err := c.Create(context.Background(), &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "debug",
```
mhrabovcin, line 139: It would be nice to have this defined as a constant with a clear name referring to its use. Reading later 

```
	defer func() {
		c.Delete(context.Background(), "debug", metav1.DeleteOptions{})
	}()
```

Wasn't clear to me what is happening.

joejulian (changes requested):
What is the goal of this PR? It looks like it's intended to be a test that reloader does what reloader does. I feel this is the wrong place for this kind of test.

This test actually tests that reloader does its job, that nginx and kubernetes work, and that we can start a kind cluster - all of which can be more succinctly tested using kuttl in kubeaddons-tests and none of which test that our opinionated configuration doesn't break anything.

reloader functionality should be tested by stakater. If that testing does not give us confidence, we should add additional tests upstream.

The default helm chart configuration should be tested in stakater-charts, again, upstream.

Only the opinionated changes we make to the upstream default configuration need to be tested in kba, but the only change we make are the log output type and the resources. Testing the log output is easy - make sure we're getting json. How do we test that our limits are within reason? Can we test this instead?

makkes:
> What is the goal of this PR? It looks like it's intended to be a test that reloader does what reloader does. I feel this is the wrong place for this kind of test.
> 
> This test actually tests that reloader does its job, that nginx and kubernetes work, and that we can start a kind cluster - all of which can be more succinctly tested using kuttl in kubeaddons-tests and none of which test that our opinionated configuration doesn't break anything.
> 
> reloader functionality should be tested by stakater. If that testing does not give us confidence, we should add additional tests upstream.
> 
> The default helm chart configuration should be tested in stakater-charts, again, upstream.
> 
> Only the opinionated changes we make to the upstream default configuration need to be tested in kba, but the only change we make are the log output type and the resources. Testing the log output is easy - make sure we're getting json. How do we test that our limits are within reason? Can we test this instead?

@joejulian @faiq @dlipovetsky please have a look at the Jira task linked in the PR's description for more context.

makkes:
@jimmidyson you have an opinion here, too?

jimmidyson:
As I understand it, this test was added to use the DKP test framework API in a real addon to get a feel for the API.

makkes:
Now for a more extensive answer to your remarks, @joejulian:

As @jimmidyson already mentioned we picked this Addon to make use of the new testing framework on a very simple Addon. I agree with you that we're basically testing reloader itself here. However, my understanding is that the Addon provides a means to install reloader with custom configuration using kubeaddons. Given that we need to make sure that the Addon still does what we expect it to do when e.g. the chart version is updated, e.g. it could happen that some chart values provided through the Addon may not play well (anymore) with a new chart version. That's where I believe such an end-to-end test still makes sense. 🤔 

wdyt?

faiq:
Hey Max I've read through the original issue listed [here](https://jira.d2iq.com/browse/D2IQ-72909) and through the [subtask](https://jira.d2iq.com/browse/D2IQ-73341) that you're working on, and I have found that the test is symptom of different ideas when it comes to testing. 

There's an open debate on _what_ we need to be testing. 

There's an open debate on _how_ we should be testing things as well. 

Both of these things require extensive discussion and buy in as an organization as a whole. We have @mesosphere/sig-ksphere-testing every two weeks for us to talk about these sort big changes that we're going to be taking on

In terms of my opinion on this, I'm not sure that we need to be writing 200+ lines of go code to provision a kind cluster, deploy kubeaddons, deploy nginx seems a lot of work to do setup for a test rather than doing testing of the actual component itself. I'd like to move more towards a model of decoupling tests from provisioning infrastructure. I could be convinced otherwise, but we should have the discussion where we can layout and discuss the merits of different methodologies first. Next week is eng-week, so hopefully we can all spend some time discussing the best way to move forward as an organization together.

Copying @orsenthil @jimmidyson @dlipovetsky @joejulian @shaneutt who've all expressed interest in discussing this topic.  

Excited to move the conversation together as an engineering organization and work to a better solution!

joejulian:
> As was already mentioned we picked this Addon to make use of the new testing framework on a very simple Addon.

If this is intended to just be a demo, please add the do not merge tag. I suspect, however, that is not the intent. I, personally, disagree with the method of testing in this instance. I'd prefer we use kuttl tests if this is a desired test.

> I agree with you that we're basically testing reloader itself here. However, my understanding is that the Addon provides a means to install reloader with custom configuration using kubeaddons.

That is true. Unfortunately, this test also doesn't test that. It tests its own json file, not the addon values. If a PR changes the addon values and neglects to change the json values, this test will no longer be valid.

> Given that we need to make sure that the Addon still does what we expect it to do when e.g. the chart version is updated, e.g. it could happen that some chart values provided through the Addon may not play well (anymore) with a new chart version. That's where I believe such an end-to-end test still makes sense. thinking

I feel that we should add tests that address changes to the values as those values are changed. We currently change two sets of values for this chart. Log output type and resource allocations. For these two changes, we would want to test that the logs continue to be output in json, and that managing an expected number of configmaps, secrets, and/or pods will not cause the reloader controller to exceed those resource limits.

If we change a different value in the future, that change should be accompanied by a test. For example, when we choose to set `serviceMonitor.enabled` to `true`, for reloader, we would add a test to validate that the metrics endpoint returns a 200 as part of that PR.

joejulian:
Oh, also a different perspective on "we need to make sure that the Addon still does what we expect it to do when e.g. the chart version is updated", we know that's what reloader does regardless of any values change we can make. We don't need to validate that it does that. On the other hand, we _do_ set the annotations on _other_ addons to trigger a reload on a configmap change. Testing _that_ addon for the desired behavior would seem more aligned with testing the values changes we make as part of our opinionated configuration.

faiq:
I've made a test here using KUDO that checks to see if reloader doesn't crash with a workload bigger than ours. Please check it out 

https://github.com/mesosphere/kubeaddons-tests/pull/11

joejulian:
Closing this in favor of the kuttl test.

PR #760, opened 2020-12-18
Remove elasticsearch requirement from fluentbit

branden:
**What type of PR is this?**

Bug

This allows fluentbit to deploy without elasticsearch enabled. If a user
would like to disable elasticsearch and use an alternative log sink,
they would otherwise have to edit the fluentbit addon spec and remove
the dependency.

* https://jira.d2iq.com/browse/D2IQ-70491

Getting this PR to a reviewable state required some investigation of the kubeaddons controller, so I included a commit that collects its logs during test cleanup. I've also updated Teamcity configuration to collect logs as build artifacts.

```release-note
[fluentbit] Allows the fluentbit addon to deploy without elasticsearch enabled.
```

**Checklist**

### DIFF
--- a/addons/elasticsearch/elasticsearch.yaml
+++ b/addons/elasticsearch/elasticsearch.yaml
@@ -10,7 +10,7 @@ metadata:
     # on the cluster, this hack will trigger re-queue on Addons until one exists.
     kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: "true"
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "6.8.2-9"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "6.8.2-10"
     appversion.kubeaddons.mesosphere.io/elasticsearch: "6.8.2"
     values.chart.helm.kubeaddons.mesosphere.io/elasticsearch: "https://raw.githubusercontent.com/mesosphere/charts/73fba37/stable/elasticsearch/values.yaml"
     # Some StatefulSet changes can't be applied, so we need to delete and recreate.
--- a/addons/elasticsearch/elasticsearch.yaml
+++ b/addons/elasticsearch/elasticsearch.yaml
@@ -35,6 +35,11 @@ spec:
     version: 1.32.0
     values: |
       ---
+      cluster:
+        config:
+          # Disable automatic creation of audit log indexes.
+          # This is intended to reject audit logs until their index template has been created. See data.hooks.postStart.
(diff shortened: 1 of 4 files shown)


gracedo:
>     # This allows us to have fluentbit wait until ES is deployed and has the right configurations up, in particular	
>     # setting up index templates

If fluentbit is installed before ES is up, would there be fluentbit config issues?

branden:
I think there's a risk that if fluent-bit deploys first then it may start sending audit log data to Elasticsearch before the audit log index template is created. There are some type inconsistencies in audit logs that force us to drop a fraction of them, since Elasticsearch doesn't support compound types for values, so the purpose of that index template is to select types that ensure we collect as many logs as possible. If Elasticsearch starts receiving logs before the index template exists then it will infer template types from the incoming data, and this could result in a greater number of dropped logs.

To see if this was a problem I deployed a couple Konvoy clusters, one with KBA pointed to master and another cluster with KBA pointed to this branch. I let them run for a while and then took a look at audit log metrics on the fluent-bit dashboard. The following screenshots are from a ~20 minute span of time at least an hour past deployment:

master:
Screen Shot 2020-12-18 at 6 59 34 PM

this PR:
Screen Shot 2020-12-18 at 6 59 56 PM

The top graph in each screenshot plots audit logs incoming to fluent-bit. The bottom graph plots audit logs sent to Elasticsearch (green) on the left axis and dropped audit logs (orange) on the right axis.

Incoming audit log rate is about the same on both clusters, but the cluster for this PR has a lower successful output rate, and about 5x the rate of dropped audit logs. It's still a small number relative to incoming logs, about 1 log dropped every 2 seconds, but that concerns me.

samba:
@branden  IIUC, the main problem here is that the _fluentbit_ addon is responsible for configuring something in _elasticsearch_ -- right? 

This sounds like we need to separate concerns. Configuration for ES needs to reside with the ES addon, and should be removed from the fluentbit addon. 

The dropped logs problem is significant: we should avoid that, and we should not ship a product with that problem. Please attempt to move the configuration to the appropriate addons, and assess whether this problem persists. 

We also cannot continue to ship with this cross-dependency in effect; customers must be allowed to use fluentbit *without* elasticsearch.

branden:
I updated the PR to make ES reject audit logs until the index template has been created, and audit log throughput looks like what I saw on master:

Screen Shot 2021-01-12 at 5 01 42 PM

s12chung (approved):

gracedo (approved):

PR #801, opened 2021-01-21
[fluent-bit] Replace dots with underscores in audit logs

branden:
**What type of PR is this?**

Bug

Audit logs may include resource labels, which may include dots (`.`). Elasticsearch treats dots as field delimiters when parsing incoming records, which can lead to conflicts when assembling the final record to store, causing it to be rejected. This PR reduces the number of dropped audit logs by replacing all periods with underscores in audit log records that are sent to Elasticsearch.

A couple screenshots to demonstrate the impact of this change on a running cluster:

Screen Shot 2021-01-20 at 6 26 27 PM

Screen Shot 2021-01-20 at 6 25 51 PM

* https://jira.d2iq.com/browse/D2IQ-73839

```release-note
\[fluent-bit\] Fixes an issue causing some audit logs to be dropped.
```

**Checklist**


s12chung (approved):

gracedo (approved):

sebbrandt87 (changes requested):
wait, let me check if we should also solve this with the filter we added to general logs.
As this change isn't good in my opinion.

alejandroEsc:
could we wait until after merging kibnana 7.9.3 please?

sebbrandt87 (commented):
unblock, but we might add more changes as discussed

sebbrandt87:
@branden yes :)
I just thought we add here the elasticsearch index_template changes that are might needed as well. But also I am okay with it being merged like it is now.

branden:
Ah sorry for the confusion @sebbrandt87. I know we discussed checking the audit log index template to see if it needs updates, but I saw that as a separate follow-on task. I want to get this PR merged as soon as possible because it solves a customer issue without (AFAICT) introducing new problems. Sounds like you're cool with this being merged as-is but LMK if you think we need to revert. I'll set up a meeting for us to review the template once the current release frenzy is over. Here's a ticket for tracking: https://jira.d2iq.com/browse/D2IQ-73958

PR #809, opened 2021-01-22
chore: split test to upgrade and install parts

alejandroEsc:
**What type of PR is this?**

CHORE

Helps testing in a more realistic way as we do not support deleting stateful addons and downgrading. 

```release-note
NONE
```

**Checklist**

### DIFF
--- a/test/addons_test.go
+++ b/test/addons_test.go
@@ -326,13 +326,13 @@ func testgroup(t *testing.T, groupname string, version string, jobs ...clusterTe
 		return err
 	}
 
-	doUpgrade, addonDeployments, err := checkIfUpgradeIsNeeded(t, groupname)
+	doUpgrade, addons, err := checkIfUpgradeIsNeeded(t, groupname)
 
 	if doUpgrade {
 		t.Logf("=== Running UPGRADE job")
 
 		t.Logf("testing upgrade group %s", groupname)
-		err = testGroupUpgrades(t, groupname, version, jobs, addonDeployments)
+		err = testGroupUpgrades(t, groupname, version, jobs, addons)
 		if err != nil {
 			return err
 		}
--- a/test/addons_test.go
+++ b/test/addons_test.go
@@ -550,6 +550,14 @@ func testGroupUpgrades(t *testing.T, groupname string, version string, jobs []cl
 		addonDefaults,
 	)
 	th.Load(addonUpgrades...)
+
+	for _, job := range jobs {
+		th.Load(testharness.Loadable{
+			Plan: testharness.DefaultPlan,
+			Jobs: testharness.Jobs{job(t, tcluster)},
+		})
+	}
+
 	th.Load(addonCleanup)
 
 	// Collect kubeaddons controller logs during cleanup.
@@ -575,13 +583,6 @@ func testGroupUpgrades(t *testing.T, groupname string, version string, jobs []cl
 			return err
 		}},
 	})
-
-	for _, job := range jobs {
-		th.Load(testharness.Loadable{
-			Plan: testharness.DefaultPlan,
-			Jobs: testharness.Jobs{job(t, tcluster)},
-		})
-	}
 
 	defer th.Cleanup()
 	th.Validate()
--- a/test/addons_test.go
+++ b/test/addons_test.go
@@ -674,6 +674,39 @@ configInline:
     addresses:
     - "172.17.1.200-172.17.1.250"
 `,
+	"elasticsearch": `
+---
+# Reduce resource limits so elasticsearch will deploy on a kind cluster with limited memory.
+client:
+  heapSize: 256m
+  resources:
+    limits:
+      cpu: 1000m
+      memory: 512Mi
+    requests:
+      cpu: 500m
+      memory: 256Mi
+master:
+  heapSize: 256m
+  resources:
+    limits:
+      cpu: 1000m
+      memory: 512Mi
+    requests:
+      cpu: 100m
+      memory: 256Mi
+data:
+  persistence:
+    size: 4Gi
+  heapSize: 1024m
+  resources:
+    limits:
+      cpu: 1000m
+      memory: 1536Mi
+    requests:
+      cpu: 100m
+      memory: 1024Mi
+`,
 	"prometheus": `
 ---
 # Remove dependency on persistent volumes and Konvoy's "etcd-certs" secret.
@@ -686,10 +719,10 @@ kubeEtcd:
 `,
 }
 
-func newCluster(groupName string, version string, node v1alpha4.Node, t *testing.T) (testcluster.Cluster, error) {
-	if groupName == "aws" || groupName == "azure" || groupName == "gcp" || groupName == allAWSGroupName || groupName == "elasticsearch" {
-		provisioner := groupName
-		if groupName == allAWSGroupName || groupName == "elasticsearch" {
+func newCluster(groupname string, version string, node v1alpha4.Node, t *testing.T) (testcluster.Cluster, error) {
+	if groupname == "aws" || groupname == "azure" || groupname == "gcp" || groupname == allAWSGroupName {
+		provisioner := groupname
+		if groupname == allAWSGroupName {
 			provisioner = "aws"
 		}
 		path, _ := os.Getwd()
--- a/test/addons_test.go
+++ b/test/addons_test.go
@@ -585,6 +585,12 @@ func testGroupUpgrades(t *testing.T, groupname string, version string, jobs []cl
 	})
 
 	defer th.Cleanup()
+
+	if len(addonUpgrades) == 0 {
+		t.Logf("INFO: NO UPGRADES to be tested, EXITING")
+		return nil
+	}
+
 	th.Validate()
 	th.Deploy()
 	th.Default()
--- a/test/addons_test.go
+++ b/test/addons_test.go
@@ -214,6 +214,65 @@ func TestAzureGroup(t *testing.T) {
 // Private Functions
 // -----------------------------------------------------------------------------
 
+func checkIfUpgradeIsNeeded(t *testing.T, groupname string) (bool, []v1beta2.AddonInterface, error) {
+	t.Logf("determining which addons in group %s need to be upgrade tested", groupname)
+
+	var doUpgrade bool
+	addons := groups[groupname]
+	for _, addon := range addons {
+		if err := overrides(addon); err != nil {
+			return false, nil, err
+		}
+	}
+
+	addonDeploymentsArray := []v1beta2.AddonInterface{}
+	for _, newAddon := range addons {
+		t.Logf("verifying whether upgrade testing is needed for addon %s", newAddon.GetName())
+		oldAddon, err := addontesters.GetLatestAddonRevisionFromLocalRepoBranch("../", comRepoRemote, comRepoRef, newAddon.GetName())
+		if err != nil {
+			if strings.Contains(err.Error(), "directory not found") {
+				t.Logf("no need to upgrade test %s, it appears to be a new addon (no previous revisions found in branch %s)", newAddon.GetName(), comRepoRef)
+				addonDeploymentsArray = append(addonDeploymentsArray, newAddon)
+				continue
+			}
+			return false, nil, err
+		}
+		if oldAddon == nil {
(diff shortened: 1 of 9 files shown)


s12chung:
```
 harness.go:73: ERROR: could not forward port to prometheus pod: pod prometheus-prometheus-kubeaddons-prom-prometheus-0 is not running, it's in phase Pending
 ```

**Comment by gracedo** [comment]
> ```
>  harness.go:73: ERROR: could not forward port to prometheus pod: pod prometheus-prometheus-kubeaddons-prom-prometheus-0 is not running, it's in phase Pending
> ```

This is an error we are also seeing in #778 , hoping increasing the sleep will fix that one

**Comment by alejandroEsc** [comment]
Seems to be an issue deploying traefik in one of the tests

File: test/addons_test.go [hunk 1]
```diff
@@ -345,6 +493,7 @@ func testgroup(t *testing.T, groupname string, version string, jobs ...clusterTe
 		if err != nil {
 			if strings.Contains(err.Error(), "directory not found") {
 				t.Logf("no need to upgrade test %s, it appears to be a new addon (no previous revisions found in branch %s)", newAddon.GetName(), comRepoRef)
+				addonDeploymentsArray = append(addonDeploymentsArray, newAddon)
```
gracedo, line 496: if no upgrade test needed, we probably don't need to actually do anything more for the addon right? this would essentially rerun `testGroupDeployment` which is just deploying the latest addon version

test/addons_test.go
```diff
@@ -366,24 +515,34 @@ func testgroup(t *testing.T, groupname string, version string, jobs ...clusterTe
...
 		}
 
-		t.Logf("found old version of addon %s %s (revision %s) and new version %s (revision %s)", newAddon.GetName(), oldRev, oldVersion, newVersion, newRev)
-		if oldVersion.GT(newVersion) {
-			return fmt.Errorf("revisions for addon %s are broken, previous revision %s is newer than current %s", newAddon.GetName(), oldVersion, newVersion)
-		}
-		if !newVersion.GT(oldVersion) {
+		if newVersion.EQ(oldVersion) {
 			t.Logf("skipping upgrade test for addon %s, it has not been updated", newAddon.GetName())
+			addonDeploymentsArray = append(addonDeploymentsArray, oldAddon)
```
gracedo, line 520: same comment here: if no upgrade test needed, we probably don't need to actually do anything more for the addon right? this would essentially rerun `testGroupDeployment` which is just deploying the latest addon version

test/addons_test.go
```diff
@@ -366,24 +515,34 @@ func testgroup(t *testing.T, groupname string, version string, jobs ...clusterTe
...
 			return err
 		}
 
+		// append old version to Deployments
+		addonDeploymentsArray = append(addonDeploymentsArray, oldAddon)
+		// append for upgrades
 		addonUpgrades = append(addonUpgrades, addonUpgrade)
 	}
 
+	addonDeployment, err := addontesters.DeployAddons(t, tcluster, addonDeploymentsArray...)
```
gracedo, line 541: I think at this point we should check if there are any addons that actually need to be upgrade-tested, if there are none then we should be able to skip running any of the steps here

test/addons_test.go
```diff
@@ -214,6 +214,65 @@ func TestAzureGroup(t *testing.T) {
...
+		} else if oldVersion.GT(newVersion) {
+			return false, nil, fmt.Errorf("revisions for addon %s are broken, previous revision %s is newer than current %s", newAddon.GetName(), oldVersion, newVersion)
+		} else {
+			t.Logf("found old version of addon %s %s (revision %s) and new version %s (revision %s)", newAddon.GetName(), oldRev, oldVersion, newVersion, newRev)
+		}
+
+		doUpgrade = true
+
+		// append old version to Deployments
+		addonDeploymentsArray = append(addonDeploymentsArray, oldAddon)
```
gracedo, line 270: We shouldn't need to deploy the oldAddon - separately deploying an oldAddon that will be upgrade-tested isn't necessary since the `UpgradeAddon` plan actually [deploys the older version prior to upgrading](https://github.com/mesosphere/kubeaddons/blob/master/test/utils/addons.go#L289-L292), so in this case if we were to deploy the oldAddon first, the `UpgradeAddon` plan would delete the addon deployed (even if it's the old version) then redeploy the old version

gracedo (commented):
Looks like finally tests passed (no upgrades), hooray!! For the upgrades test, which won't be triggered in this PR, I think overall we should be able to avoid running through two loops of addons and instead rework `checkIfUpgradeIsNeeded` to include all the logic. I hope my comments made sense, as I'm sure we all are aware, this test code is a bit hard to understand :) happy to go over this over zoom tomorrow if something doesn't make sense! Very possible I also may have missed something between this test code and the kubeaddons test utils code 😅

test/addons_test.go
```diff
@@ -366,24 +578,37 @@ func testgroup(t *testing.T, groupname string, version string, jobs ...clusterTe
...
 		t.Logf("INFO: addon %s was modified and will be upgrade tested", newAddon.GetName())
+
 		addonUpgrade, err := addontesters.UpgradeAddon(t, tcluster, oldAddon, newAddon)
 		if err != nil {
 			return err
 		}
 
+		// append for upgrades
 		addonUpgrades = append(addonUpgrades, addonUpgrade)
 	}
```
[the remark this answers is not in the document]
alejandroEsc: That is something i tried to do, however, the issue is that the cluster reference is passed inito each upgrade object, and that reference is created on creation of the cluster. I could not seperate these concerns without actually changing the harness code.

test/addons_test.go
```diff
@@ -214,6 +214,65 @@ func TestAzureGroup(t *testing.T) {
 // Private Functions
 // -----------------------------------------------------------------------------
 
+func checkIfUpgradeIsNeeded(t *testing.T, groupname string) (bool, []v1beta2.AddonInterface, error) {
```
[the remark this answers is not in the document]
alejandroEsc: my comment above, this boolean for now allows us to get what addons to deploy first and (which some will get upgraded afte) (so a bit of a efficiency boos theer) and keeps a cluster from being unnecessarily deployed.

test/addons_test.go
```diff
@@ -214,6 +214,65 @@ func TestAzureGroup(t *testing.T) {
...
+			if strings.Contains(err.Error(), "directory not found") {
+				t.Logf("no need to upgrade test %s, it appears to be a new addon (no previous revisions found in branch %s)", newAddon.GetName(), comRepoRef)
+				addonDeploymentsArray = append(addonDeploymentsArray, newAddon)
+				continue
+			}
+			return false, nil, err
+		}
+		if oldAddon == nil {
+			t.Logf("no need to upgrade test %s, it appears to be a new addon (no previous revisions found in branch %s)", newAddon.GetName(), comRepoRef)
+			continue // new addon, upgrade test not needed
```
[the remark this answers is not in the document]
alejandroEsc: Yes you are right!

test/addons_test.go
```diff
@@ -214,6 +214,65 @@ func TestAzureGroup(t *testing.T) {
...
+			return false, nil, err
+		}
+
+		if newVersion.EQ(oldVersion) {
+			t.Logf("skipping upgrade test for addon %s, it has not been updated", newAddon.GetName())
+			addonDeploymentsArray = append(addonDeploymentsArray, oldAddon)
+			continue
+		} else if oldVersion.GT(newVersion) {
+			return false, nil, fmt.Errorf("revisions for addon %s are broken, previous revision %s is newer than current %s", newAddon.GetName(), oldVersion, newVersion)
+		} else {
```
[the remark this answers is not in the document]
alejandroEsc: agreed

test/addons_test.go
alejandroEsc: I was missing this, thank you! We should only deploy those addons that are not upgraded, we do have to test the group after all.

gracedo (approved):
LGTM! I wonder if we want to run a dummy upgrade test on this PR before merging just to make sure the upgrade tests work? Maybe just bumping the revision on an addon, seeing that tests pass then reverting the change before merging?

alejandroEsc:
> LGTM! I wonder if we want to run a dummy upgrade test on this PR before merging just to make sure the upgrade tests work? Maybe just bumping the revision on an addon, seeing that tests pass then reverting the change before merging?

That sounds good, let me branch from this and create a new pr but with a change.

test/addons_test.go
```diff
@@ -260,6 +315,35 @@ func cleanupNodeVolumes(numberVolumes int, nodePrefix string, node *v1alpha4.Nod
...
+
+	t.Logf("=== Running INSTALL job")
+
+	// TODO these happen sequentially, we should do so in parallel when we have enough confidence
+	err = testGroupDeployment(t, groupname, version, jobs)
+	if err != nil {
+		return err
+	}
+
+	doUpgrade, addonDeployments, err := checkIfUpgradeIsNeeded(t, groupname)
```
s12chung, line 329: prefer calling this `addons`, as it's making me think of the old `addonDeployments`, but it's a nit

test/addons_test.go
```diff
@@ -214,6 +214,61 @@ func TestAzureGroup(t *testing.T) {
...
+			return false, nil, err
+		}
+
+		if newVersion.EQ(oldVersion) {
+			t.Logf("skipping upgrade test for addon %s, it has not been updated", newAddon.GetName())
+			addonDeploymentsArray = append(addonDeploymentsArray, oldAddon)
+			continue
+		} else if oldVersion.GT(newVersion) {
+			return false, nil, fmt.Errorf("revisions for addon %s are broken, previous revision %s is newer than current %s", newAddon.GetName(), oldVersion, newVersion)
+		}
```
s12chung, line 264: i wonder why we need to do this comparison, shouldn't the controller handle it....

s12chung (approved):
overall, looks fine to me.

test/addons_test.go
alejandroEsc: hmmm, yeah i see what you mean, we can still make that change.

test/addons_test.go
alejandroEsc: It is for the sake of creating addonUpgrades objects, which has little to do how kubernetes handles resources.

test/addons_test.go
```diff
@@ -214,6 +214,61 @@ func TestAzureGroup(t *testing.T) {
...
+
+		doUpgrade = true
```
gracedo, line 266: oops, sorry I meant the `else` wasn't needed to wrap the log but let's keep the log in! (https://github.com/mesosphere/kubernetes-base-addons/pull/809/commits/5da80d493f3257b938d403c8ad42bff58af846b7#diff-6b1d50a89d4ac455215d0f0969ed4e5507e3a91528e4ef6a20e37d6d515781b2L263-L264)
```suggestion
		t.Logf("found old version of addon %s %s (revision %s) and new version %s (revision %s)", newAddon.GetName(), oldRev, oldVersion, newVersion, newRev)
		doUpgrade = true
```

test/addons_test.go
```diff
@@ -366,24 +574,37 @@ func testgroup(t *testing.T, groupname string, version string, jobs ...clusterTe
...
+
+	if len(addonUpgrades) == 0 {
+		t.Logf("INFO: NO UPGRADES to be tested, EXITING")
+		close(stop)
+		return nil
+	}
+
 	th := testharness.NewSimpleTestHarness(t)
 	th.Load(
 		addontesters.ValidateAddons(addons...),
```
gracedo, line 610: Can't comment on line 612 but below with `addonDefaults` plan (terribly named 🙄 maybe we should rename it to `waitForAddons` or something clearer lol) - This is waiting for all the addons including addons that haven't been deployed yet (ones that will be upgrade-tested, as they are deployed in the Upgrade plan) which is why we're seeing the tests fail in #817 . I think we should be passing in `deployments...` instead. `addonDefaults` is defined at [line 543](https://github.com/mesosphere/kubernetes-base-addons/pull/809/files#diff-6b1d50a89d4ac455215d0f0969ed4e5507e3a91528e4ef6a20e37d6d515781b2R543) (have to expand the diff to see)

alejandroEsc:
closing in favor of PR #817

PR #718, opened 2020-12-04
bug(fluentbit): ES mapping correction / modify

sebbrandt87:
**What type of PR is this?**

Bug

**What this PR does/ why we need it**:

- We are adding the filters here to have the workaround for labels 
kubernetes.label.app=foobar.
Elasticsearch cannot accept strings on a field and beneath it new 
objects.
So it will reject all new entries when there is any 
kubernetes.label.app=foobar (string)
entry already present in an index. So we remap / modify the 
kubernetes.label.app to be kubernetes.label.app.kubernetes.io/name. (see https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)
-- adapted from https://github.com/fluent/fluent-bit/issues/483#issuecomment-593393980

- also updated to latest fluent-bit chart and version 1.6.8

**Which issue(s) this PR fixes**:

jql=key in (D2IQ-73455)

[D2IQ-73455]

[D2IQ-73455]: https://jira.d2iq.com/browse/D2IQ-73455

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
- modify `kubernetes.label.app` to `kubernetes.label.app.kubernetes.io/name`
- Fluent Bit 1.6.8
```

**Checklist**

* [x] The commit message explains the changes and why are needed.
* [x] The code builds and passes lint/style checks locally.
* [ ] The relevant subset of integration tests pass locally.
* [ ] The core changes are covered by tests.
* [ ] The documentation is updated where needed.

PR #814, opened 2021-01-25
fix(fluentbit): Added filters for audit.*

sebbrandt87:
**What type of PR is this?**

Bug

- Added filters for audit.* in the same manner as
for kube.* See #718
All to resolve the conflict of common old usage of
labels. New common labels see
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Closes: #801

jql=key in (D2IQ-73839)

[D2IQ-73839]

[D2IQ-73839]: https://jira.d2iq.com/browse/D2IQ-73839

After this we need to take care of the kubeaddon `elasticsearch`
index_template for `kubernetes-audit`, as this isn't valid anymore. 
Must be adapted or dropped overall.

```release-note
- Audit log label field `app`, rewrite to `app.kubernetes.io/name`
```

**Checklist**


branden (changes requested):
This PR doesn't satisfy the requirements of https://jira.d2iq.com/browse/D2IQ-73839. Rewriting a set of labels we predict isn't sufficient. A customer may apply any arbitrary label with a dot in it to a resource. Thus, we need to make sure that _any_ label with a dot in it does not cause Elasticsearch to drop audit logs for that resource. This was my intention with #801. For what it's worth, I launched a konvoy cluster using the KBA from this PR and it is dropping more audit logs than #801 (contrast with the graph in the #801 PR description, which sometimes reaches 0):

Screen Shot 2021-01-25 at 12 42 55 PM

sebbrandt87:
It drops because of the "wrong" / no longer valid index_template. But we can discuess this also in the Zoom call.

sebbrandt87:
All righty, we talked, and agreed on #718
We just maybe define some more additions to other parts.

PR #816, opened 2021-01-26
Automated chart bump gatekeeper-0.5.1

d2iq-dispatch:
**What type of PR is this?**
Chore

upgrade gatekeeper addon to newer version

https://jira.d2iq.com/browse/D2IQ-73506

The upstream chart is located at https://github.com/open-policy-agent/gatekeeper/tree/master/charts/gatekeeper

I have done the manual upgrade test. Here are the steps that I followed:
```
1. Provisioned a Konvoy 1.6 cluster with gatekeeper enabled
2. Created a gatekeeper policy to validate allowed repository that a pod can use
     ( Here is the configuration that I used to test https://github.com/arbhoj/gatekeeper-policies/tree/master/allowed-repos-deployment-example)
3. Validated that the policy is working 
4. Upgraded gatekeeper using this chart
5. Validated that the policy is still working
```

```release-note

```

**Checklist**

* [ ] *If a chart is changed, the chart version is correctly incremented.*

### DIFF
--- a/addons/gatekeeper/gatekeeper.yaml
+++ b/addons/gatekeeper/gatekeeper.yaml
@@ -1,4 +1,3 @@
----
 apiVersion: kubeaddons.mesosphere.io/v1beta1
 kind: Addon
 metadata:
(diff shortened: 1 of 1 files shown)


s12chung (approved):

addons/gatekeeper/gatekeeper.yaml
```diff
@@ -34,15 +33,10 @@ spec:
...
+      replicas: 2
       image:
-        tag: v3.0.4-beta.1
-      resources:
-        limits:
-          cpu: 500m
-          memory: 500Mi
-        requests:
-          cpu: 200m
-          memory: 300Mi
```
joejulian, line 48: Unless these are the new chart defaults, we include these as our opinionated choices.

joejulian (changes requested):

addons/gatekeeper/gatekeeper.yaml
GoelDeepak: There are new chart defaults https://github.com/mesosphere/charts/blob/master/staging/gatekeeper/values.yaml#L56. I am open-minded if we want to keep these vs chart defaults. I don't have enough data point about how we arrived at those values vs upstream. In the absence of that data point, I was favoring upstream.

alejandroEsc: these were probably arrived at via MWT. I would suggest we keep these values for now and dial them up/down based on the next mwt run.

hectorj2f: The values were extracted by our learnings to keep gatekeeper running healthy while doing load tests.

hectorj2f: It seems the code of gatekeeper changed now. We are not going to exercise it during this new MWT. So we can keep the defaulted ones in the chart. The gatekeeper initial resource allocation were lower before, so we experienced OOM issues during our MWT tests.

GoelDeepak: Sounds good.

joejulian (approved):

PR #829, opened 2021-01-28
kibana: Downgrade to 6.8.10 to fix regression

gracedo:
**What type of PR is this?**

bug

downgrade to 6.8.10 to fix regression - metrics were not getting exported

https://jira.d2iq.com/browse/D2IQ-73949

```release-note
fix: Downgrade kibana and elasticsearch to 6.8.10 to fix a regression
```

**Checklist**

### DIFF
--- a/addons/elasticsearch/elasticsearch.yaml
+++ b/addons/elasticsearch/elasticsearch.yaml
@@ -10,8 +10,8 @@ metadata:
     # on the cluster, this hack will trigger re-queue on Addons until one exists.
     kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: "true"
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "6.8.13-11"
-    appversion.kubeaddons.mesosphere.io/elasticsearch: "6.8.13"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "6.8.13-12"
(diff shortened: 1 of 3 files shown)


hectorj2f (approved):

hectorj2f (commented):
I am wondering if there would be any issue with ES being 6.8.13. I believe it won't.

gracedo:
> I am wondering if there would be any issue with ES being 6.8.13. I believe it won't.

oh good point, I remember @alejandroEsc having to match the versions. I would expect the tests to fail if not compatible though, and they're passing here. will wait for @alejandroEsc to approve

alejandroEsc:
> > I am wondering if there would be any issue with ES being 6.8.13. I believe it won't.
> 
> oh good point, I remember @alejandroEsc having to match the versions. I would expect the tests to fail if not compatible though, and they're passing here. will wait for @alejandroEsc to approve

Kibana will always complain if mismatch. So expect lots of logs.

gracedo:
> > > I am wondering if there would be any issue with ES being 6.8.13. I believe it won't.
> > 
> > 
> > oh good point, I remember @alejandroEsc having to match the versions. I would expect the tests to fail if not compatible though, and they're passing here. will wait for @alejandroEsc to approve
> 
> Kibana will always complain if mismatch. So expect lots of logs.

to be safe I'll change ES to match :)

samvantran (approved):

branden (approved):

PR #856, opened 2021-02-04
fix: addons test for release branches

makkes:
**What type of PR is this?**

Bug

**What this PR does/ why we need it**:

The addons tests assumed they were always run with `master` being the
target branch. However, during release, we often target different
branches (e.g. release/3.2). Therefore, the tests need to be
configurable.

Now, the tests can be configured with a command-line flag:

```
go test -tags experimental -timeout 60m -race -v -run TestPrometheusGroup -kba-branch release/3.2
```

This would run the prometheus test group tests against the release/3.2
branch.

If the flag is not provided then `master` is assumed to be the target
branch so the default behaviour doesn't change.

After this commit has been merged, we need to do 2 things:

1. backport to release/3.2 branch
2. change step 2 of this build configuration to pass the target branch
   to the tests:
   https://teamcity.mesosphere.io/admin/editBuildRunners.html?id=buildType:kubeaddons_KubernetesBaseAddons

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
none
```

**Checklist**

* [ ] The commit message explains the changes and why are needed.
* [ ] The code builds and passes lint/style checks locally.
* [ ] The relevant subset of integration tests pass locally.
* [ ] The core changes are covered by tests.
* [ ] The documentation is updated where needed.

PR #858, opened 2021-02-04
[3.2] fix: addons test for release branches

makkes:
backport of #856 

**What type of PR is this?**

Bug

The addons tests assumed they were always run with `master` being the
target branch. However, during release, we often target different
branches (e.g. release/3.2). Therefore, the tests need to be
configurable.

Now, the tests can be configured with a command-line flag:

```
go test -tags experimental -timeout 60m -race -v -run TestPrometheusGroup -kba-branch release/3.2
```

This would run the prometheus test group tests against the release/3.2
branch.

If the flag is not provided then `master` is assumed to be the target
branch so the default behaviour doesn't change.

```release-note
none
```

### DIFF
--- a/test/addons_test.go
+++ b/test/addons_test.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


makkes:
The Kubernetes Base Addons (sig-ksphere-catalog) check will likely fail because TC isn't configured to provide the right parameters, yet.

joejulian (approved):

armandgrillet (approved):

PR #1054, opened 2021-04-14
[release/2] prometheus: use existing storagespec

gracedo:
**What type of PR is this?**

bug

If either prometheus or alertmanager resources exist and have a storage
attribute defined in their spec, reuse that instead of what's defined
in the values.

This prevents the accidental redefinition of the volumeClaimTemplate and
the abandonment of old data.

https://jira.d2iq.com/browse/COPS-6870

```release-note
prometheus: Fixed an upgrade issue that occurred when upgrading from a prior chart version to a version >=9.3.0 and <=9.3.7 which had a bug where the desired PVC name prefix was being discarded. This bug caused upgrades to that version to create and use new storage claims instead of the existing ones. This feature will now always use the storage setting defined in the cluster, if there is one, reusing the existing claims. (COPS-6870)
```

**Checklist**

### DIFF
--- a/addons/prometheus/prometheus.yaml
+++ b/addons/prometheus/prometheus.yaml
@@ -10,7 +10,7 @@ metadata:
     # on the cluster, this hack will trigger re-queue on Addons until one exists.
     kubeaddons.mesosphere.io/hack-requires-defaultstorageclass: "true"
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "0.38.1-19"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "0.38.1-20"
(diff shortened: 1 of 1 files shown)


gracedo:
No more 2.x milestones so maybe that means we're not cutting any more release/2 cuts since it'll be soon unsupported?

joejulian:
Unless you think we have to.

gracedo:
I think this is probably low-priority since it fixes the upgrade from <9.3.0 to 9.3.x. I'm not sure how many customers, if any, we have that have yet to upgrade to konvoy 1.5

joejulian:
Closing per discussion.

PR #1128, opened 2021-05-12
[auto-bump][chart] vsphere-csi-driver-1.3.0

d2iq-dispatch:
**What type of PR is this?**

Feature

Bump [vsphere CSI to 2.2.0](https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/master/docs/book/releases/v2.2.0.md).

[Ran e2e tests with these changes](https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_Konvoy_KonvoyE2eVsphere/3117932?buildTab=tests&status=passed).

https://jira.d2iq.com/browse/D2IQ-75438

1. I pasted the upstream manifests in the first commit.
2. Added back the Helm templating.
3. Minor reverts to fix e2e failures.

```release-note
Update vpshere-csi-driver to v2.2.0 to pull in the fix for duplicate operations failure when attaching PVs (https://github.com/kubernetes-sigs/vsphere-csi-driver/issues/580) (COPS-6906)
```

**Checklist**

* [x] *If a chart is changed, the chart version is correctly incremented.*

### DIFF
--- a/addons/vsphere-csi-driver/vsphere-csi-driver.yaml
+++ b/addons/vsphere-csi-driver/vsphere-csi-driver.yaml
@@ -1,4 +1,3 @@
----
 apiVersion: kubeaddons.mesosphere.io/v1beta1
 kind: ClusterAddon
 metadata:
@@ -7,9 +6,9 @@ metadata:
     kubeaddons.mesosphere.io/name: vsphere-csi-driver
     kubeaddons.mesosphere.io/provides: csi-driver
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "1.2.1-1"
-    appversion.kubeaddons.mesosphere.io/vsphere-csi-driver: "2.0.1"
-    values.chart.helm.kubeaddons.mesosphere.io/vsphere-csi-driver: "https://raw.githubusercontent.com/mesosphere/charts/5565912/stable/vsphere-csi-driver/values.yaml"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "2.2.0-1"
+    appversion.kubeaddons.mesosphere.io/vsphere-csi-driver: "2.2.0"
+    values.chart.helm.kubeaddons.mesosphere.io/vsphere-csi-driver: "https://raw.githubusercontent.com/mesosphere/charts/ee4430b/stable/vsphere-csi-driver/values.yaml"
 spec:
   requires:
     - matchLabels:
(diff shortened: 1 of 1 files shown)


samvantran:
Updated the release note to reference the COPS and issue number (more descriptive reason why we're bumping the addon). 

Question though - should PR title say we've updated the chart version or the appversion? IMHO it should be appversion (2.2.0) but @joejulian @dkoshkin I leave you to decide.

Also this is the automated duplicate of #1127 so whichever you want to keep, please use this release note

joejulian:
Bumped... But why? What problem does this solve? What value does this add?

Let's add value by being descriptive rather than make our customers dig in to it themselves.

samvantran:
I see your point. We could offer more upfront but IMO I think customers are still going to want to dig into the problem themselves? IIRC customers actually have COPS numbers that Support will then go and report back if the upgrade fixes their problem. Either way someone will read into it?

But I updated it some more bc you're right, helps to provide more context upfront. They can read the issue more if they are interested. How is it now?

addons/vsphere-csi-driver/vsphere-csi-driver.yaml
```diff
@@ -7,9 +6,9 @@ metadata:
     kubeaddons.mesosphere.io/name: vsphere-csi-driver
     kubeaddons.mesosphere.io/provides: csi-driver
   annotations:
-    catalog.kubeaddons.mesosphere.io/addon-revision: "1.2.1-1"
-    appversion.kubeaddons.mesosphere.io/vsphere-csi-driver: "2.0.1"
-    values.chart.helm.kubeaddons.mesosphere.io/vsphere-csi-driver: "https://raw.githubusercontent.com/mesosphere/charts/5565912/stable/vsphere-csi-driver/values.yaml"
+    catalog.kubeaddons.mesosphere.io/addon-revision: "2.2.0-1"
```
faiq, line 9: Is this the right version?

faiq (changes requested):
Blocked until i can figure out what the version is supposed to be.

addons/vsphere-csi-driver/vsphere-csi-driver.yaml
dkoshkin: Oh good catch shouldn't this come from https://github.com/mesosphere/charts/blob/a41acbd59fdb11635e8f8119f33308d17045cbde/stable/vsphere-csi-driver/Chart.yaml#L7?

